PR #10, opened 2020-03-11
Adding win_zip module along with integration tests

jborean93:
_From @JimFicarra on Jan 19, 2020 20:07_

##### SUMMARY

Addition of win_zip module to support zipping and to complement the win_unzip. Initial support is to allow zipping of a file or a directory recursively but can be expanded upon.  There is an inherent 2GB file size limit on each source file in .NET and PowerShell (which relies on .NET underneath)

It was not possible to support COM Shell.Application as the least common denominator (as in win_unzip) because the API call required UI interaction that did not have options to suppress.  As such a design decision was made to support PowerShell first, and .NET 4.5 or higher second.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

win_zip

##### ADDITIONAL INFORMATION

- Zips a single file into a zip file, or a directory (recursively) into a zip file.

- Options include 'creates' for idempotence as well as 'delete_src' to remove the source file/directory after zipping.

- Check Mode is supported

Please note that as I am submitting this as a Comcast Employee, internal approval/process was required to submit this.  As such, I am required to submit this verbiage with the PR:

> We are only granting a copyright license, and no other licenses, to all users and developers of the project, present and future, pursuant to the license of the project.

```paste below

```

_Copied from original issue: ansible/ansible#66608_

### DIFF
--- a/plugins/modules/win_zip.ps1
+++ b/plugins/modules/win_zip.ps1
@@ -1,0 +1,253 @@
+#!powershell
+
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+#Requires -Module Ansible.ModuleUtils.Legacy
+#Requires -Version 5.0
+
+Set-StrictMode -Version 2.0
+
+$ErrorActionPreference = "Stop"
+#TODO/Future Enhancements
+#-------------------------
+# - support recurse or just content for dirs
+# - support multiple source files and dirs and globs
+# - "force" option to overwrite existing zip file
+# - Provide an 'update' parameter to add or remove files from the zip
+# - Support other archive types - $pcx_extensions = @('.bz2', '.gz', '.msu', '.tar', '.zip')
+
+# Notes
+# -----
+# The operations should be atomic - the zip shouldn't be created if anything fails
+
+# Initialize it here as it's used for the failifempty
+$result = @{
+    changed = $false
+    source_deleted = $false
+}
+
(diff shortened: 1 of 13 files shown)


jborean93:
_From @rnsc on Jan 21, 2020 09:45_

@jborean93 Isn't the `#Requires -Module Ansible.ModuleUtils.Legacy` not recommended anymore for new Windows modules?

The new modules should use `#AnsibleRequires -CSharpUtil Ansible.Basic` right?

jborean93:
_From @nitzmahone on Jan 21, 2020 22:32_

Tough call, but I'd be -1 to adopt this one without content idempotence. The `creates` UI pattern is a bit of a trip-hazard (other than the "force re-archiving" case, when would it have a value different than `dest`?), and it would probably need to be deprecated if/when true idempotence got implemented. The "delete_src" case is also unfriendly to idempotence (unless it were altered to ignore missing source in that case, but that's a little scary too). 

A zip module can be a good thing, but IMO most folks would expect true idempotence from the module (or at least a UI that could accommodate it easily in the future). If we wanted to accept it without archive-level idempotence, I'd suggest dropping `creates` and `delete_src`, and maybe adding a `force_rearchive` or something if an analogue to `creates: ~` is wanted.

jborean93:
Trying to clean up the remaining migrated PRs. Due to the age I am going to close this PR but if someone still wishes to push forward with the changes here they are more than welcome to open a new PR.

PR #12, opened 2020-03-11
Add new module win_psscript for installing scripts from PSRepositories

jborean93:
_From @briantist on Feb 23, 2020 03:34_

##### SUMMARY

Adding new module `win_psscript` as a counterpart to `win_psmodule`.

Most common options and scenarios accounted for, including mix/max/required versions, add/remove and update to latest.

Includes tests.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

win_psscript

##### RELATED PRs:
- #67499
- #67594

_Copied from original issue: ansible/ansible#67674_

### DIFF
--- a/plugins/modules/win_psscript.ps1
+++ b/plugins/modules/win_psscript.ps1
@@ -146,7 +146,8 @@ if ($state -eq 'absent') {
             if (-not $module.CheckMode) {
                 $existing.Values | Uninstall-Script -Force -ErrorAction Stop
             }
-        } catch {
+        }
+        catch {
             $module.FailJson("Error uninstalling scripts.", $_)
         }
     }
@@ -154,7 +155,8 @@ if ($state -eq 'absent') {
 else { # state is 'present' or 'latest'
     try {
         $remote = Find-Script @pFind -ErrorAction Stop
-    } catch {
+    }
+    catch {
         $module.FailJson("Error searching for scripts.", $_)
     }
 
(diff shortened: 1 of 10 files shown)


jborean93:
_From @briantist on Feb 23, 2020 21:11_

Thanks @jhawkesworth , I didn't put real scripts in mostly because the only publicly published script I personally use is one that I published a few weeks ago and it probably wouldn't have wide interest (it's related to JEA). Most of the scripts I use are internal to my company and published in their private repository, so I actually didn't have a whole lot of real-world examples. 

I took a look at the popular scripts in PSGallery and changed 2 of the examples to use real ones that seem like they're relatively safe and might be of interest, but if you know of any that a good fit I'd be happy to replace the existing examples with some that are more useful!

jborean93:
_From @jhawkesworth on Feb 24, 2020 18:17_

Thanks @briantist  only script I've used from PSG is the AnsibleVault one.  Changes look good to me.

jborean93:
_From @briantist on Feb 25, 2020 01:11_

> Thanks @briantist only script I've used from PSG is the AnsibleVault one. Changes look good to me.

Oh nice, I took a look at that but it's a module so it would be a good example for `win_psmodule`.

jborean93:
Just had to remove the `version_added` entry in the docs but the rest should be good to go. Will merge once it is green.

jborean93:
Unfortunately looks like CI is broken, will have to look further into this next week when I am back online.

jborean93:
The original issue was caused by recent change to ansible-devel and has since been reverted https://github.com/ansible/ansible/issues/68530. The 2nd issue was there were still modules calling `ansible.windows` modules without the prefix, have since fixed those and CI is green again.

briantist:
Great, thanks for the fixups on this!

PR #56, opened 2020-03-22
win_psrepository - add script and publish options, resolve locations

### LINKED ISSUE #48: Add script and publish location support to win_psrepository
_From @briantist on Feb 17, 2020 21:31_

##### SUMMARY

The `win_psrepository` module only supports the `source` (module source) location, which means that the repositories it registers can't be used for downloading scripts, or publishing either modules or scripts.

Ideally it would add the remaining 3 location types so that complete repositories can be used.

The other 3 (PowerShell names) are:
- `PublishLocation`
- `ScriptSourceLocation`
- `ScriptPublishLocation`

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

lib/ansible/modules/windows/win_psrepository.ps1

##### ADDITIONAL INFORMATION

Example usage:

```yaml
win_psrepository:
  name: MyRepo
  script_source_location: https://myrepo.local/scripts
  publish_location: https://myrepo.local/module_publish
  script_publish_location: https://myrepo.local/script_publish
```

_Copied from original issue: ansible/ansible#67501_

briantist:
_Manual migration of https://github.com/ansible/ansible/pull/67499_
Original changelog fragment not included here since I don't see where it would go in the new repo.

----
##### SUMMARY

* Add parameters for script and publish locations ( Fixes #48 )

* Allow for forced update
  * Normal mode of operation on existing repositories is to update only the settings you supplied (same as previously)
  * `force` will make the repository match your settings exactly, by unregistering and re-registering the repository if there are changes
  * some desired changes can't be made without `force`, like if you wanted to "unset" a publish location after having previously set it to any value; there's no way to do that without registering anew.
* Add new tests for above.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

 lib/ansible/modules/windows/win_psrepository.ps1

### DIFF
--- a/plugins/modules/win_psrepository.ps1
+++ b/plugins/modules/win_psrepository.ps1
@@ -1,10 +1,14 @@
 #!powershell
 
+# Copyright: (c) 2020, Brian Scholer <@briantist>
 # Copyright: (c) 2018, Wojciech Sciesinski <wojciech[at]sciesinski[dot]net>
 # Copyright: (c) 2017, Daniele Lazzari <lazzari@mailup.com>
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
 #Requires -Module Ansible.ModuleUtils.Legacy
+
+[System.Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSUseDeclaredVarsMoreThanAssignments', '', Justification='Some vars are referenced via Get-Variable')]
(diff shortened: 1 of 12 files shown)


briantist:
I added a stub for `prepare_http_tests` which allows the tests to directly use `httpbin.org` so they run for now. 

Added a separate issue and PR for reintroducing `prepare_http_tests`:
- #57 
- #58

PR #62, opened 2020-03-30
New module win_psmodule_info

briantist:
##### SUMMARY

This is a new info module for PowerShell modules (not related to Ansible modules). It returns detailed information about modules on the system, and adds in info from PowerShellGet relevant to modules that were installed from a repository.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

win_psmodule_info

##### ADDITIONAL INFORMATION

From examples:
```yaml
- name: Get info about all modules on the system
  community.windows.win_psmodule_info:

- name: Get info about the ScheduledTasks module
  community.windows.win_psmodule_info:
    name: ScheduledTasks

- name: Get info about networking modules
  community.windows.win_psmodule_info:
    name: Net*

- name: Get info about all modules installed from the PSGallery repository
  community.windows.win_psmodule_info:
    repository: PSGallery
  register: gallery_modules

- name: Update all modules retrieved from above example
  community.windows.win_psmodule:
    name: "{{ item }}"
    state: latest
  loop: "{{ gallery_modules.modules | map(attribute=name) }}"

- name: Get info about all modules on the system
  community.windows.win_psmodule_info:
  register: all_modules

- name: Find modules installed from a repository that isn't registered now
  set_fact:
    missing_repository_modules: "{{
      all_modules
      | json_query('modules[?repository!=null && repository==repository_source_location].{name: name, version: version, repository: repository}')
      | list
    }}"

- debug:
    var: missing_repository_modules
```


briantist:
ready_for_review

plugins/modules/win_psmodule_info.py
```diff
@@ -0,0 +1,439 @@
...
+          description:
+            - The maximum version of the module.
+          type: str
+          sample: '2.9'
+        required_version:
+          description:
+            - The exact version of the module required.
+          type: str
+          sample: '3.1.4'
+    nested_modules:
```
jborean93, line 354: A quick check on this it seems like `nested_modules` is just a dict in the format

```
nested_modules:
  Count: x
  Value:
  - name: abc
    ...
  - name: def
    ...
```

Seems like we need to specially handle whatever this type is in .NET so it's actually serialized as a list.

plugins/modules/win_psmodule_info.py
```diff
@@ -0,0 +1,439 @@
...
+    power_shell_version:
+      description:
+        - The minimum version of PowerShell that the module requires.
+      type: str
+      sample: '5.1'
+    prefix:
+      description:
+        - The default prefix applied to C(Verb-Noun) commands exported from the module, resulting in C(Verb-PrefixNoun) names.
+      type: str
+    private_data:
```
jborean93, line 399: I'm not sure if this is actually what the value should be but `private_data` in some of the nested modules of `NetAdapter` are showing up as

```
private_data:
  CmdletOverObjects:
    ClassName: ROOT/StandardCimv2/MSFT_....
    CmdletAdpater: Microsoft.PowerShell.Cmdletization.Cim.CimCmdletAdapter
```

I think is is correct but probably just something to look over to make sure.

briantist: Yeah that's expected. I actually handle `private_data` specially specifically for non-Script/Manifest modules because of this. From the comment on lines 93-97:

> PrivateData can contain any data but a module manifest is a static file that can't contain references or problematic types like [Type]. Unfortunately some module types like CIM (and presumably binary?) seem to be able to populate that with whatever they want.
> 
> As a precaution then, for module type that is not Script or Manifest, we limit PrivateData to a shallow depth.

It was those nested CIM modules that brought it to my attention; they had `[Type]` values which balloon serialization really quickly :(

plugins/modules/win_psmodule_info.py
briantist: Did you test this from within ansible or by running it directly in PowerShell?

I got those values outside of Ansible, but not within it, strangely enough. I had a workaround prepared and everything and ended up removing it

It would probably wouldn't hurt anything to put it back; basically old redundant type data that's added has to be removed:
https://stackoverflow.com/a/48858780/3905079

plugins/modules/win_psmodule_info.py
jborean93: Thanks for confirming

plugins/modules/win_psmodule_info.py
jborean93: It was run within Ansible.

briantist: Ok I'm going to put that workaround back in then since it doesn't seem to hurt anything.

Weirdly it didn't serialize like that in shippable either (the tests would have failed), I wonder what difference it would make being run from Ansible sometimes vs others.

briantist: @jborean93 pushed the change, let's see if that fixes the serialization for you

jborean93: Looks good, coming up as a list of dicts for me now.

PR #67, opened 2020-04-03
win_pagefile does nothing when override is false

LukasGibb:
##### SUMMARY

win_pagefile does nothing when override is false

Fixes https://github.com/ansible/ansible/issues/68652

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

win_pagefile

##### ADDITIONAL INFORMATION

The powershell command used gave unexpected results (always True) causing the module to never apply changes as it thought the page file was system managed.

See full info here: https://github.com/ansible/ansible/issues/68652

### DIFF
--- a/plugins/modules/win_pagefile.ps1
+++ b/plugins/modules/win_pagefile.ps1
@@ -130,11 +130,11 @@ if ($state -eq "absent") {
         $result.changed = $true
     }else
     {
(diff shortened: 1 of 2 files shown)


jborean93:
Sorry for taking too long to get to this PR. Just rerunning the tests to make sure it's still good with the recent changes made to it.

PR #79, opened 2020-04-21
New Module - win_pssession_configuration

briantist:
##### SUMMARY

Adding the `win_pssession_configuration` module and associated action plugin, for configuring PS remoting sessions.

Supports most options you'll find in `New-PSSessionConfigurationFile` and `Register-PSSessionConfiguration`.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

win_pssession_configuration

##### ADDITIONAL INFORMATION
Updating these configurations restarts the WinRM service, which causes an unrecoverable failure for Ansible. Connection errors can't be handled with `rescue:` and `async:` is not enough on its own to work around this, as the initial setup of the async task doesn't complete in time to avoid the interruption. Further, because unreachable hosts can't be handled in `rescue:`, it would be difficult to reliably use `async_status` in a play.

The action plugin works through this with a few tricks. First, if we're in check mode, then `async` is not used. In testing I found that it was possible, but it's unneeded and will take longer.

When not in check mode, the module is always run async, and then `async_status` is used internally to give a consistent non-async experience. Because of the aforementioned connection/unreachable difficulties though, we catch exceptions and on an exception, we go directly to calling `wait_for_connection` to wait out any interruption and then try again.

Also of note is that in testing, the `-NoServiceRestart` parameter on the `*-PSSessionConfiguration` commands does not work. It does change some of the warnings and verbose output it shows but it does not actually stop the service from being restarted. As a result that option is not included in this module and isn't used internally. Instead we assume any change will cause a service restart.

### DIFF
--- a/plugins/action/win_pssession_configuration.py
+++ b/plugins/action/win_pssession_configuration.py
@@ -70,6 +70,8 @@ class ActionModule(ActionBase):
             '_async_dir': ntpath.dirname(status['results_file'])
         }
 
+        max_retries = 3
+        retries = 0
         while not check_mode:
             try:
                 # check up on the async job
@@ -91,6 +93,11 @@ class ActionModule(ActionBase):
                 time.sleep(self._task.poll)
 
             except BaseException as e:
+                retries += 1
+                if retries >= max_retries:
+                    display.vvvv("Max retries reached.")
+                    raise e
+                display.vvvv("Retrying (%s of %s)" % (retries, max_retries))
                 display.vvvv("Falling back to wait_for_connection: %r" % e)
                 wait_connection_action.run(task_vars=task_vars)
 
--- a/plugins/action/win_pssession_configuration.py
+++ b/plugins/action/win_pssession_configuration.py
@@ -25,9 +25,9 @@ class ActionModule(ActionBase):
     def run(self, tmp=None, task_vars=None):
         self._supports_check_mode = True
         self._supports_async = False
-        check_mode      = self._play_context.check_mode
-        async_timeout   = self._task.args.get('async_timeout', self._default_async_timeout)
-        async_poll      = self._task.args.get('async_poll', self._default_async_poll)
+        check_mode = self._play_context.check_mode
+        async_timeout = self._task.args.get('async_timeout', self._default_async_timeout)
+        async_poll = self._task.args.get('async_poll', self._default_async_poll)
 
         result = super(ActionModule, self).run(tmp, task_vars)
 
--- a/plugins/action/win_pssession_configuration.py
+++ b/plugins/action/win_pssession_configuration.py
@@ -30,7 +30,7 @@ class ActionModule(ActionBase):
 
         # fake out the super so it doesn't stop us from using check mode with async
         if check_mode:
-            self._task.async_val = None
+            self._task.async_val = 0
 
         result = super(ActionModule, self).run(tmp, task_vars)
 
--- a/plugins/action/win_pssession_configuration.py
+++ b/plugins/action/win_pssession_configuration.py
@@ -32,10 +32,6 @@ class ActionModule(ActionBase):
         if check_mode:
             self._task.async_val = None
 
-        # if I don't do this it seems to fail in module_manifest.py, line 301, in _create_powershell_wrapper
-        if self._task.async_val is None:
-            self._task.async_val = 0
-
         result = super(ActionModule, self).run(tmp, task_vars)
 
         # build the wait_for_connection object for later use
@@ -57,6 +53,11 @@ class ActionModule(ActionBase):
         # if it's not in check mode, call the module async so the WinRM restart doesn't kill ansible
         if not check_mode:
             self._task.async_val = async_timeout or self._default_async_timeout
+
+        # if I don't do this it seems to fail in module_manifest.py, line 301, in _create_powershell_wrapper
+        #  but only in shippable 🤔
+        if self._task.async_val is None:
+            self._task.async_val = 0
 
         result = status = self._execute_module(
             task_vars=task_vars,
--- a/plugins/action/win_pssession_configuration.py
+++ b/plugins/action/win_pssession_configuration.py
@@ -55,7 +55,7 @@ class ActionModule(ActionBase):
             self._task.async_val = async_timeout or self._default_async_timeout
 
         # if I don't do this it seems to fail in module_manifest.py, line 301, in _create_powershell_wrapper
-        #  but only in shippable 🤔
+        # but only in shippable
         if self._task.async_val is None:
             self._task.async_val = 0
 
@@ -70,7 +70,7 @@ class ActionModule(ActionBase):
             return result
 
         # remove the async flag so that our calls to async_status and wait_for_connection don't try to run async
-        self._task.async_val = None
+        self._task.async_val = 0
 
         async_status_args = {
             'jid': status['ansible_job_id'],
--- a/plugins/action/win_pssession_configuration.py
+++ b/plugins/action/win_pssession_configuration.py
@@ -19,6 +19,7 @@ def clean_async_result(reference_keys, obj):
         obj.pop(key)
     return obj
 
+
 class ActionModule(ActionBase):
     _default_async_timeout = 300
     _default_async_poll = 1
--- a/plugins/action/win_pssession_configuration.py
+++ b/plugins/action/win_pssession_configuration.py
@@ -14,6 +14,10 @@ from ansible.playbook.task import Task
 from ansible.utils.display import Display
 display = Display()
 
+def clean_async_result(reference_keys, obj):
+    for key in reference_keys:
+        obj.pop(key)
+    return obj
 
 class ActionModule(ActionBase):
     _default_async_timeout = 300
@@ -135,4 +139,4 @@ class ActionModule(ActionBase):
             # let's swallow errors during implicit cleanup to aovid interrupting what was otherwise a successful run
             display.vvvv("Clean up of async status failed on the remote host: %r" % e)
 
-        return result
+        return clean_async_result(status.keys(), result)
--- a/plugins/action/win_pssession_configuration.py
+++ b/plugins/action/win_pssession_configuration.py
@@ -24,22 +24,18 @@ class ActionModule(ActionBase):
 
     def run(self, tmp=None, task_vars=None):
         self._supports_check_mode = True
-        self._supports_async = True
-        check_mode = self._play_context.check_mode
-        async_timeout = self._task.async_val
-        async_poll = self._task.poll
-
-        # fake out the super so it doesn't stop us from using check mode with async
-        if check_mode:
-            self._task.async_val = 0
+        self._supports_async = False
+        check_mode      = self._play_context.check_mode
+        async_timeout   = self._task.args.get('async_timeout', self._default_async_timeout)
+        async_poll      = self._task.args.get('async_poll', self._default_async_poll)
 
         result = super(ActionModule, self).run(tmp, task_vars)
 
         # build the wait_for_connection object for later use
         wait_for_connection_task = self._task.copy()
         wait_for_connection_task.args = {
-            'timeout': async_timeout or self._default_async_timeout,
-            'sleep': async_poll or self._default_async_poll
+            'timeout': async_timeout,
+            'sleep': async_poll,
         }
         wait_connection_action = self._shared_loader_obj.action_loader.get(
             'wait_for_connection',
@@ -53,7 +49,7 @@ class ActionModule(ActionBase):
 
         # if it's not in check mode, call the module async so the WinRM restart doesn't kill ansible
         if not check_mode:
-            self._task.async_val = async_timeout or self._default_async_timeout
+            self._task.async_val = async_timeout
 
         result = status = self._execute_module(
             task_vars=task_vars,
--- a/plugins/action/win_pssession_configuration.py
+++ b/plugins/action/win_pssession_configuration.py
@@ -30,6 +30,10 @@ class ActionModule(ActionBase):
 
         # fake out the super so it doesn't stop us from using check mode with async
         if check_mode:
+            self._task.async_val = None
+
+        # if I don't do this it seems to fail in module_manifest.py, line 301, in _create_powershell_wrapper
+        if self._task.async_val is None:
             self._task.async_val = 0
 
         result = super(ActionModule, self).run(tmp, task_vars)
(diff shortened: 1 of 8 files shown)


plugins/action/win_pssession_configuration.py
```diff
@@ -0,0 +1,110 @@
...
+        # Retries here is a fallback in case the module fails in an unexpected way
+        # which can sometimes not properly set the failed field in the return.
+        # It is not related to async retries.
+        # Without this, that situation would cause an infinite loop.
+        max_retries = 3
+        retries = 0
+        while not check_mode:
+            try:
+                # check up on the async job
+                job_status = self._execute_module(
```
jborean93, line 84: Why are you calling the async_status module and not the action plugin like you've done for `wait_for_connection`?

plugins/modules/win_pssession_configuration.ps1
```diff
@@ -0,0 +1,551 @@
...
+    # in the config file options, and then re-register later if needed.
+    # But if the RunAs credential is wrong, the register will fail, and since we already removed
+    # the existing one, it will be gone.
+    #
+    # So let's ensure we can actually use the credential by running something (whoami.exe) with the become
+    # module util, that way we can fail before touching the existing config.
+    if ($opt_session.Contains('RunAsCredential')) {
+        $cred = $opt_session.RunAsCredential
+        try {
+            $result = [Ansible.Become.BecomeUtil]::CreateProcessAsUser($cred.Username, $cred.GetNetworkCredential().Password, 'whoami.exe')
```
jborean93, line 500: Instead of using `Ansible.Become` to test a credential you should use `[Ansible.AccessToken]::LogonUser('username', 'domain', 'password', 'Interactive', 'Default')` to just logon. I can't think of a reason to actually create a process with those credentials.

Have a look at [win_user](https://github.com/ansible-collections/ansible.windows/blob/1c36f86276bea6c7a5ac4b178bd93626378da93a/plugins/modules/win_user.ps1#L42-L70) to see how it tests credentials. I'm not sure if you wanted to ignore any particular error codes like it does or considering everything that wasn't a successful logon as a failure.

plugins/modules/win_pssession_configuration.ps1
```diff
@@ -0,0 +1,551 @@
...
+            $module.Debug("RunAsCredential check whoami.exe output: $($result.StandardOut)")
+            if ($result.ExitCode -ne 0) {
+                throw
+            }
+        }
+        catch {
+            $module.FailJson('Could not validate RunAs Credential.')
```
jborean93, line 507: It would be good to include some error information for both the user's and developers perspective

```suggestion
            $module.FailJson('Could not validate RunAs Credential: $($_.Exception.Message)', $_)
```

plugins/modules/win_pssession_configuration.ps1
```diff
@@ -0,0 +1,551 @@
...
+            $psso = $opt_session
+            # Wait-WinRMConnection
+            Set-PSSessionConfiguration @psso
+        }
+    }
+}
+catch [System.Management.Automation.ParameterBindingException] {
+    $e = $_
+    if ($e.Exception.ErrorId -eq 'NamedParameterNotFound') {
+        $psv = $PSVersionTable.PSVersion.ToString(2)
```
jborean93, line 534: Nice I didn't know `[Version]` had this ability.

plugins/modules/win_pssession_configuration.ps1
```diff
@@ -0,0 +1,551 @@
...
+        $param = $e.Exception.ParameterName
+        $cmd = $e.InvocationInfo.MyCommand.Name
+        $message = "Parameter '$param' is not available for '$cmd' in PowerShell $psv."
+    }
+    else {
+        $message = "Unknown parameter binding error."
```
jborean93, line 540:
```suggestion
        $message = "Unknown parameter binding error: $($e.Exception.Message)"
```

tests/integration/targets/win_pssession_configuration/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
+# This file is part of Ansible
+
+# Copyright: (c) 2020, Brian Scholer <@briantist>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+---
+- block:
+    - ansible.windows.setup:
```
jborean93, line 7: Gathering facts can be an expensive operation and it looks like you just want to get the PowerShell version. I recommend just turning this to a `win_shell` task to get the `$PSVersion` to save on testing time.

plugins/modules/win_pssession_configuration.ps1
briantist: Perfect! I was looking for something like this; I'll switch it over. I don't truly have a need to start any process.

plugins/modules/win_pssession_configuration.ps1
briantist: Me neither! I saw the `-1`'s in testing and figured I'd check out the overloads.

plugins/action/win_pssession_configuration.py
briantist: Hm, I wrote the action plugin in a much more basic version of this module from last year, and I don't remember why that's the case. It might be that I wrote that part first, as it's much more obvious how to call modules than other action plugins. I'll try it out the other way.

plugins/action/win_pssession_configuration.py
```diff
@@ -0,0 +1,110 @@
...
+    _default_async_poll = 1
+
+    def __init__(self, *args, **kwargs):
+        super(ActionModule, self).__init__(*args, **kwargs)
+
+    def run(self, tmp=None, task_vars=None):
+        self._supports_check_mode = True
+        self._supports_async = True
+        check_mode = self._play_context.check_mode
+        async_timeout = self._task.async_val
```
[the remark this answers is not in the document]
briantist: Yeah, sort of. A lot of manipulation of the "real" `async` and `poll` values has to be done, so this is to save the value passed in.

It does also get used later, as the real `async` value for the module invocation. In that way, I think letting the user use the standard `async` keyword is the most intuitive. 

With a separate module option, it would be a bit confusing if a user used the keyword, or used both the module option and keyword.

Since I need to do some manual setting of the attributes on `self._task` anyway I can't really avoid touching those fields.

If it's not a dealbreaker for you I'd prefer to keep it as is and flow through user-specified `async` and `poll` keyword values.

tests/integration/targets/win_pssession_configuration/tasks/main.yml
briantist: Does using `gather_subset` change that? By excluding everything else it seems to be running very fast for me.
No big deal to change it to `win_shell` just wondering if `gather_subset` is actually saving any execution time or just filtering out after the fact.

briantist: Took a look at the `setup.ps1` code and it seems like it skips all the expensive stuff the way I'm calling it, which seems to agree with the performance I'm seeing (github isn't showing the `gather_subset` option when I view this as a comment on the issue conversation so you might have missed it).

If you'd prefer I'd change it to `win_shell` anyway let me know.

plugins/action/win_pssession_configuration.py
briantist: I figured it out why I couldn't get this to work in the past; I was trying to duplicate the current task object when I should have been creating a new one. I found a good example in `task_executor.py` which set me straight; thanks!

plugins/action/win_pssession_configuration.py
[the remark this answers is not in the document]
jborean93: I would strong urge you against it. If we ever start supporting running action plugins asynchronously and not just the module part then this module will never work as you've co-opted the `async` and `poll` task directives for use inside the action plugin and not towards it. While you probably can't avoid touching `self._task` entirely it should be a end goal to avoid using it as much as you can.

It may even be more beneficial for you to run the module in the action plugin without restarting the WinRM service, execute the service restart and then wait until the service is back up and running similar to how `win_reboot` works instead of trying to run things entirely through async but I don't think that is necessary right now.

tests/integration/targets/win_pssession_configuration/tasks/main.yml
jborean93: If it works then let's just leave it be. `gather_subset` will definitely reduce the time it takes to run especially if you are only getting certain facts.

plugins/action/win_pssession_configuration.py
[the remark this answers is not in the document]
briantist: I can't avoid restarting the WinRM service because the cmdlets that modify the session configurations don't properly support it. They have a switch for not restarting the service but it doesn't actually work, and the service is restarted anyway (I originally approached the problem that way as it would have been much cleaner, but unfortunately it's not possible).

Is there a proposal/PR for making actions async? It might be possible to support both depending on how its implemented.

Anyway I'll look into changing these values into a module option instead for now. They'll have to be used in the action plugin in basically the same way as now, but if it makes for better future proofing I'll go that route.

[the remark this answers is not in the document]
jborean93: > I can't avoid restarting the WinRM service because the cmdlets that modify the session configurations don't properly support it

The module could still set things up and return a command that needs to be run for `Register-PSSessionConfiguration` and the "service restart" step is just running that particular command.

> Is there a proposal/PR for making actions async? It might be possible to support both depending on how its implemented.

No proposal but it should be something to think about before trying to reuse an existing task directive(s) for an internal purpose.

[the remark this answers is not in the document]
briantist: > The module could still set things up and return a command that needs to be run for `Register-PSSessionConfiguration` and the "service restart" step is just running that particular command.

Hm that's interesting. How would you recommend running the command itself? Invoking `raw` / `win_shell` basically and then catching the exception and doing `wait_for_connection` maybe?

Sidestepping the need for async at all does sound attractive. 

I'll look into doing that vs. switching away from the keywords toward module options; might want to do that in a future revision if I end up running into unexpected issues, but I like the idea a lot.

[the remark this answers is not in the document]
briantist: I started looking into that idea but realized it would be a little more involved than I initially thought, so I've gone the other way of adding async control options to the module.

To help alleviate incompatibility with future async enhancements, the module now reports that it doesn't support async, so that should solve any weirdness in trying to use the async keywords along with or instead of the new module options (it will fail).

The documentation has been updated to reflect this, and there are a few new tests to validate the async options and that the async keywords fail as expected.

briantist:
The test failures are from httptester. I added #85 to complement #82 and address that.

plugins/modules/win_pssession_configuration.py
```diff
@@ -0,0 +1,394 @@
...
+    description:
+      - Sets a delay in seconds between each check of the asynchronous execution status.
+      - Replicates the functionality of the C(poll) keyword.
+      - Has no effect in check mode.
+    type: int
+    default: 1
+notes:
+  - This module will restart the WinRM service on any change. This will terminate all WinRM connections including those by other Ansible runs.
+  - Internally this module uses C(async) when not in check mode to ensure things goes smoothly when restarting the WinRM service.
+  - The standard C(async) and C(poll) keywords cannot be used; instead use the I(async_timeout) and I(async_poll) options to controll asynchronous execution.
```
jborean93, line 307: control*

plugins/modules/win_pssession_configuration.py
```diff
@@ -0,0 +1,394 @@
...
+  - Setting I(poll=0) will return a result that can be used with C(async_status).
```
jborean93, line 308: Is this `async_poll` now?

jborean93:
Testing works fine for me, there might be 3 things you could potentially look into but I don't want to hold up the PR for them.

* Potentially look at returning the GUID that was autogenerated by the module. In check mode we can just return `00000000-0000-0000-0000-000000000000` (I think we do that in another module).
* The return values contain some async specific info like `ansible_async_watchdog_pid`, `ansible_job_id`, `finished`, `results_file`, we should look at putting those results in a sub dict like `{"_module_async": { ... }}` instead of at the root to avoid future collision or just omit them entirely
* By default async job files are not deleted once complete, we could run `async_status: mode=cleanup` to remove the result file once it's not needed. Without this there will be a build-up of async files in `~/.ansible_async`

The reason why `async_status` does not clean up the file is so subsequent `async_status` invocations will find the job info and it can also be checked in separate processes by just the ansible_job_id.

briantist:
> Testing works fine for me, there might be 3 things you could potentially look into but I don't want to hold up the PR for them.
> 
> * Potentially look at returning the GUID that was autogenerated by the module. In check mode we can just return `00000000-0000-0000-0000-000000000000` (I think we do that in another module).

GUID is a strange property for PSSession Configurations because it turns out it's entirely optional and isn't used by anything. It's a property only of the session configuration file (which is itself optional). Even hen present, there's no way to do anything with it; none of the commands take is as a parameter (some return it with other conf file properties if it exists). If I had to guess it's just an artifact of using a common file format with other PowerShell stuff.

This module does generate if it also is generating a configuration file (because it's required for the file), but it only creates the file if it has to.

So in short, it's not very useful info to return. Happy to reconsider that if you know of a good use for it.

> * The return values contain some async specific info like `ansible_async_watchdog_pid`, `ansible_job_id`, `finished`, `results_file`, we should look at putting those results in a sub dict like `{"_module_async": { ... }}` instead of at the root to avoid future collision or just omit them entirely

I've gone ahead with just removing them, with the one exception of when the module is called with `async_poll=0` to support end-user controlled checking of async status; in that case the fields are preserved.

> * By default async job files are not deleted once complete, we could run `async_status: mode=cleanup` to remove the result file once it's not needed. Without this there will be a build-up of async files in `~/.ansible_async`
> 

This is a great idea; I added this in and it now cleans up after itself, with the one exception of using `async_poll=0` (so that the user can still check on the job themself).

It also exposed a regression whereby I was calling the remaining action plugins with async still enabled. That _seemed _to still work fine before, but it caused very strange problems when trying to call `async_status mode=cleanup` . The cleanup call worked correctly as seen on the server, but somehow, something was failing _after_ the `return` in the `run` method, saying it couldn't find the async job anymore.

I can't really explain it but I think that by having async set in this action plugin, somehow the caller was using that as a cue to do something with the async result, on the remote host.

In any case I put back a line I cleaned up too aggressively that sets the async value back to 0 after the actual module is called, and that cleared it up.

jborean93:
Cool I wasn't aware of how useless the GUID was, happy to leave that in the background as it is today.

The async changes are a ok with me, I'll do some last minute testing today and will hopefully merge it in.

Issue #83, opened 2020-04-23
win_psrepository_info - unstable test


jborean93:
##### SUMMARY
The tests for win_psrepository_info have been marked as unstable with https://github.com/ansible-collections/community.windows/pull/82. The httptester fails to be set up properly in CI for this group. I'm unable to replicate this locally without manually killing the powershell script that keeps this running in the background but this should not be happening in CI at all.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
win_psrepository_version

jborean93:
This was fixed with https://github.com/ansible-collections/community.windows/pull/128. Turns out one of the IIS tests didn't remove IIS at the end causing a conflict on port 80. This broke our httptester as any connection to `localhost:80` would go to our IIS site which didn't have the same bindings.

PR #85, opened 2020-04-30
Add unstable for win_psrepository_info

briantist:
##### SUMMARY

Issue described in #83 

Similar to PR #82 but for `win_psrepository_info`.

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/tests/integration/targets/win_psrepository/aliases
+++ b/tests/integration/targets/win_psrepository/aliases
@@ -1,3 +1,2 @@
 shippable/windows/group2
 needs/httptester
-unstable
--- a/tests/integration/targets/win_psrepository_info/aliases
+++ b/tests/integration/targets/win_psrepository_info/aliases
@@ -1,2 +1,3 @@
 shippable/windows/group3
 needs/httptester
+unstable


jborean93:
Sigh I thought I did this with https://github.com/ansible-collections/community.windows/pull/82 but I put it in the wrong target. Are you able to remove `unstable` from `win_psrepository` in this PR?

briantist:
yeah no problem, one minute

briantist:
hey I noticed something making this change. In `win_psrepository_info` I didn't use `setup_remote_tmp_dir`, whereas I did in `win_psrepository`.

It didn't seem to cause any problems initially. Any reason why that might affect httptester?

jborean93:
I don't think so as the error happening here is it failing to access the actual HTTP tester. When I was testing it the ports were definitely forwarded and the hosts entry was set up properly but something was stopping the host from accessing the httptester site. I could never narrow it down as it ran fine by itself and sometimes worked with others.

Ultimately I just didn't have time to investigate it further so decided to mark it as unstable (although I failed at that).

briantist:
The module names and test structures are so similar, easy mistake to make 😅

jborean93:
I don't think it's a problem with the group it is in, we can even see that in this CI run it ran perfectly fine. There's some sort of problem with some of the other tests that are running or maybe just when it runs in the group that's a problem. Setting it to unstable will just allow the nightly CI to be green but if someone makes a change to the module it will still run in CI.

PR #86, opened 2020-05-06
Stop using EAP Ignore

jborean93:
##### SUMMARY
Using `-ErrorAction Ignore` can have problems in advanced functions whereas `-ErrorAction SilentlyContinue` does work. This isn't usually a problem if the function succeeds but if there is a failure and it writes to the error stream (say the repo doesn't exist)  then it applies the ErrorAction and Ignore isn't supported for `$ErrorActionPreference`.

While this has been fixed in PowerShell 7 we need to ensure our code works with v3+ https://github.com/PowerShell/PowerShell/issues/1759.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_psmodule
win_psrepository

##### ADDITIONAL INFORMATION
To test this out manually in PowerShell run the following script

```powershell
Function Test-Function {
    [CmdletBinding()]
    param ([Switch]$Error)

    if ($Error) {
        Write-Error -Message "error message"
    }
}

Test-Function

Test-Function -Error

Test-Function -ErrorAction Ignore

Test-Function -Error -ErrorAction Ignore
```

On PowerShell < 7 we get

```powershell
PS C:\Users\vagrant-domain> Test-Function
PS C:\Users\vagrant-domain>
PS C:\Users\vagrant-domain> Test-Function -Error
Test-Function : error message
At line:1 char:1
+ Test-Function -Error
+ ~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Write-Error], WriteErrorException
    + FullyQualifiedErrorId : Microsoft.PowerShell.Commands.WriteErrorException,Test-Function

PS C:\Users\vagrant-domain>
PS C:\Users\vagrant-domain> Test-Function -ErrorAction Ignore
PS C:\Users\vagrant-domain>
PS C:\Users\vagrant-domain> Test-Function -Error -ErrorAction Ignore
Write-Error : The value Ignore is not supported for an ActionPreference variable. The provided value should be used
only as a value for a preference parameter, and has been replaced by the default value. For more information, see the
Help topic, "about_Preference_Variables."
At line:6 char:9
+         Write-Error -Message "error message"
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Write-Error], NotSupportedException
    + FullyQualifiedErrorId : System.NotSupportedException,Microsoft.PowerShell.Commands.WriteErrorCommand
```

On PowerShell 7+ we get

```
PS C:\Users\vagrant-domain> Test-Function
PS C:\Users\vagrant-domain>
PS C:\Users\vagrant-domain> Test-Function -Error
Test-Function: error message
PS C:\Users\vagrant-domain>
PS C:\Users\vagrant-domain> Test-Function -ErrorAction Ignore
PS C:\Users\vagrant-domain>
PS C:\Users\vagrant-domain> Test-Function -Error -ErrorAction Ignore
PS C:\Users\vagrant-domain>
```

Where this is relevant is because we were using code like `Get-PSRepository -Name repo -ErrorAction Ignore` where that repo may or may not have existed. If it did not exist then we wanted to ignore the error and continue on but due to this PowerShell bug it failed with this error.

### DIFF
--- a/plugins/modules/win_psmodule.ps1
+++ b/plugins/modules/win_psmodule.ps1
@@ -162,7 +162,7 @@ Function Add-DefinedParameter {
     )
 
     ForEach ($ParameterName in $ParametersNames) {
-        $ParameterVariable = Get-Variable -Name $ParameterName -ErrorAction Ignore
+        $ParameterVariable = Get-Variable -Name $ParameterName -ErrorAction SilentlyContinue
         if ( $ParameterVariable.Value -and $Hashtable.Keys -notcontains $ParameterName ){
                 $Hashtable.Add($ParameterName,$ParameterVariable.Value)
         }
@@ -381,7 +381,7 @@ if ( ($state -eq "latest") -and
 }
 
 if ( $repo -and (-not $url) ) {
-    $RepositoryExists = Get-PSRepository -Name $repo -ErrorAction Ignore
+    $RepositoryExists = Get-PSRepository -Name $repo -ErrorAction SilentlyContinue
     if ( $null -eq $RepositoryExists) {
         $ErrorMessage = "The repository $repo doesn't exist."
         Fail-Json $result $ErrorMessage
--- a/plugins/modules/win_psmodule.ps1
+++ b/plugins/modules/win_psmodule.ps1
@@ -27,6 +27,18 @@ $result = @{changed = $false
             nuget_changed = $false
             repository_changed = $false}
 
+
+# Enable TLS1.1/TLS1.2 if they're available but disabled (eg. .NET 4.5)
+$security_protocols = [System.Net.ServicePointManager]::SecurityProtocol -bor [System.Net.SecurityProtocolType]::SystemDefault
+if ([System.Net.SecurityProtocolType].GetMember("Tls11").Count -gt 0) {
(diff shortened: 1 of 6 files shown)


jborean93:
So this has fixed that error that masked the real problem. 2012, 2012 R2, and 2016 are failing to check PSGallery with the error

```json
{
"changed": false,
"msg": "Cant find the module ansible-test1: PowerShell Gallery is currently unavailable.  Please try again later.",
"nuget_changed": false,
"output": "",
"repository_changed": false
}
```

Technically it shouldn't be checking PSGallery as that has been removed from the repo list but PowerShellGet at particular versions seem to always check the API.

Will need to investigate this further.

jborean93:
Looks like the PSGallery result was being cached and on older hosts the default TLS policies forced the command to fail. I'm still unsure why that was an issue there but by making sure we enable TLS 1.2 the test start working again.

jborean93:
This will be backported to Ansible 2.9 and 2.8 with the following PRs in ansible/ansible

* stable-2.8 - https://github.com/ansible/ansible/pull/69362
* stable-2.9 - https://github.com/ansible/ansible/pull/69361

Issue #47, opened 2020-03-11
Add password to win_unzip module


jborean93:
_From @moleszek on Feb 11, 2020 16:15_

##### SUMMARY
According to Ansible documentation, `win_unzip` module does not support the `password` parameter.
This makes ansible unable to unpack the `.zip` file, which is password protected.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
win_unzip:
  password: "password123"

##### ADDITIONAL INFORMATION
Adding this feature will solve problems with unpacking password-protected files.

```yaml
- name: Unzip zip file
  win_unzip:
    src: C:\test\file.zip
    dest: C:\test\unpacked\
    recurse: no
    password: "password123"
```

_Copied from original issue: ansible/ansible#67308_

jborean93:
_From @midgetworrior on Feb 13, 2020 08:26_

!component +lib/ansible/modules/windows/win_unzip.ps1

jborean93:
_From @samdoran on Feb 25, 2020 15:13_

!component =lib/ansible/modules/windows/win_unzip.py

sakar97:
Is it reliable to use a 7Zip archive library to add a password parameter feature?

PR #90, opened 2020-05-20
New Feature Password in Win_unzip module

sakar97:
##### SUMMARY

#47  
Added a new feature of password parameter which unarchive password protected zip files.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

password:
    - Unarchive password-protected zip files of different extensions such as .7z,.zip

##### ADDITIONAL INFORMATION

- name: Unzip password protected files or directories
  hosts: all
  gather_facts: no
  tasks:
    - name: Unzip .7z file which is password encrypted
      win_unzip:
         src: C:\Downloads\ApplicationLogs.7z
         dest: C:\Application\Logs
         recurse: yes
         password: "abcd"
         delete_archive: yes

```paste below

```


plugins/modules/win_unzip.ps1
```diff
@@ -144,11 +145,18 @@ If ($ext -eq ".zip" -And $recurse -eq $false) {
...
+                        $password= ConvertTo-SecureString $password -AsPlainText -Force
+                        Expand-Archive -Path $src -OutputPath $dest -Password $password -WhatIf:$check_mode
+                   }
+                Else{
+                        Expand-Archive -Path $src -OutputPath $dest -Force -WhatIf:$check_mode
+              }
+	}
+	Catch{
+ 		Fail-Json -obj $result -message "Error expanding '$src' to '$dest'! Msg: $($_.Exception.Message)"
+	     }
```
jborean93, line 159: You can use splatting to your advantage here and only have to call `Expand-Archive` once. Also try to be consistent with the existing indent spacing and style that exists in the module.

```suggestion
    $expand_params = @{
        Path = $src
        OutputPath = $dest
        WhatIf = $check_mode
    }
    if ($null -ne $password) {
        $expand_params.Password = ConvertTo-SecureString -String $password -AsPlainText -Force
    }
    Try {
        Expand-Archive @expand_params
    }
    Catch {
        Fail-Json -obj $result -message "Error expanding '$src' to '$dest'! Msg: $($_.Exception.Message)"
    }
```

plugins/modules/win_unzip.py
```diff
@@ -91,6 +96,19 @@
...
+  hosts: all
+  gather_facts: no
+  tasks:
+    - name: Unzip .7z file which is password encrypted
+      win_unzip:
+         src: C:\Downloads\ApplicationLogs.7z
+         dest: C:\Application\Logs
+         recurse: yes
+         password: "abcd"
+         delete_archive: yes
```
jborean93, line 111: * No need to have the comment, the `- name:` part should explain what the example is for
    * Comments are only really useful if you need to explain the example a bit more than what `- name:` can reasonably show
* We try to avoid showing the playbook part in an example, usually we just have the task part
* Try to keep the indentation consistent with the other examples, in this case the module options are indented by 2 spaces and not 4
* As part of the move to collections the examples should be in the fully qualified collection name format
    * I'm planning on updating the existing examples of this doc and other module docs sometime shortly but new contributions should include the FQCN

```suggestion
- name: Unzip .7z file which is password encrypted
  community.windows.win_unzip:
    src: C:\Downloads\ApplicationLogs.7z
    dest: C:\Application\Logs
    recurse: yes
    password: "abcd"
    delete_archive: yes
```

plugins/modules/win_unzip.py
```diff
@@ -49,6 +49,11 @@
     description:
       - If this file or directory exists the specified src will not be extracted.
     type: path
+  password:
+    description:
+      - If a zip file is encrypted with Password
```
jborean93, line 54: Entries should be a proper sentance which ends with a full stop

```suggestion
   - If a zip file is encrypted with Password.
```

plugins/modules/win_unzip.py
```diff
@@ -49,6 +49,11 @@
...
+      - Passing a value to a password parameter requires the PSCX module to be installed.
+      - recurse must be yes
```
jborean93, line 56: Entries should be a proper sentance which ends which is capitalised and ends with a full stop

```suggestion
      - The I(recurse) option must be C(yes).
```

Curious as to why the Recurse option must be yes. The code doesn't seem to be setting that when calling `Expand-Archive`?

plugins/modules/win_unzip.py
```diff
@@ -49,6 +49,11 @@
     description:
       - If this file or directory exists the specified src will not be extracted.
     type: path
+  password:
+    description:
+      - If a zip file is encrypted with password.
+      - Passing a value to a password parameter requires the PSCX module to be installed.
+      - The I(recurse) option must be C(yes).
```
jborean93, line 56: So this is the only part that I'm unsure as to why it is needed? Is `recurse: yes` only required so it goes the PSCX path. If that's the case you should add a conditional to https://github.com/ansible-collections/community.windows/blob/2c6d6dbc6654b4d3634f0f67907d2265147d382b/plugins/modules/win_unzip.ps1#L106 to be

```
If ($ext -eq ".zip" -And $recurse -eq $false -and -not $password) {
```

This means that if the `password` is set then it will never try and extract it using the builtin tools and will immediately use PSCX for that.

jborean93:
We are definitely close, the last remaining issue is the `recurse: yes` scenario. Right now you've changed the logic to use a password for a single zip file, there's also extra calls to `Expand-Archive` at https://github.com/ansible-collections/community.windows/blob/2c6d6dbc6654b4d3634f0f67907d2265147d382b/plugins/modules/win_unzip.ps1#L166.

What you can do is remove `Path` from the `expand_params` splat/hashtable then change the following

```powershell
# https://github.com/ansible-collections/community.windows/blob/2c6d6dbc6654b4d3634f0f67907d2265147d382b/plugins/modules/win_unzip.ps1#L157
Expand-Archive -Path $src @expand_params

# https://github.com/ansible-collections/community.windows/blob/2c6d6dbc6654b4d3634f0f67907d2265147d382b/plugins/modules/win_unzip.ps1#L166
Expand-Archive -Path $_.FullName -Force @expand_params
```

This way the common parameters like `-WhatIf`, `-OutputPath`, `-Password` are all set for calls to `Expand-Archive` and only the context specific ones like `-Path` are set on the call itself.

plugins/modules/win_unzip.py
sakar97: Yes, I made `recurse: yes` only so it goes to the PSCX path and skips normal build-in tools. As we can see the doc part it clearly mentions that if you set `recurse: yes` then you need PSCX module to be installed on your target machine.  We can add the above condition which runs only when `password` is set.

jborean93: Yep, because the builtin tools with Windows don't seem to support extracting a password protected zip we can make sure we always ensure that when `password` is set it goes the PSCX route removing the need for `recurse: yes` to also be set.

jborean93:
Thanks for working on this @sakar97

PR #91, opened 2020-05-27
Update galaxy.yml

gundalow:
##### SUMMARY
Used to publish https://galaxy.ansible.com/community/windows

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -1,13 +1,13 @@
 namespace: community
 name: windows
-version: 0.0.1-beta.1
+version: 0.0.1
 readme: README.md
 authors:
 - Jordan Borean @jborean93
 - Matt Davis @nitzmahone
 description: |
   Ansible collection for community Windows pluigns.
-license_file: LICENSE
+license_file: COPYING
 tags:
 - windows
 dependencies:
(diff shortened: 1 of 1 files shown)


galaxy.yml
```diff
@@ -1,20 +1,20 @@
 namespace: community
 name: windows
-version: 0.0.1-beta.1
+version: 0.0.1
```
jborean93, line 3: I would prefer to do a beta release first when the 2.10 beta is out and only do the non-preview release when 2.10 is out officially.

gundalow: Ah, I'd already published this to Galaxy. We needed to get a full complement of collections out there.

People understand that for `0.` that these aren't for production use.

gundalow: Ah, I'd already published this to Galaxy. We needed to get a full complement of collections out there.

People understand that for `0.` that these aren't for production use.

jborean93: If you've already published that's fine, it just means that we now need to handle version_added, changelogs, and all other release based stuff which isn't really in place right now.

gundalow: There's no expectation that `0.y.x` releases on Galaxy are in any way correct or usable. So I hope by my publishing what we had yesterday doesn't impact anyone.

PR #92, opened 2020-06-02
New module: win_scoop

### LINKED ISSUE #50: Ansible module to manage packages using scoop
##### SUMMARY
Similar to how there is a `win_chocolatey` module (and related `win_chocolatey_*` modules) to manage Chocolatey packages, I would like a `win_scoop` module to manage [Scoop](https://scoop.sh/) packages.

Scoop is an alternative package manager for Windows. Check out the [GitHub repo](https://github.com/lukesampson/scoop) for more info
 
##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
win_scoop

##### ADDITIONAL INFORMATION
I've already created a small proof-of-concept at [JamieMagee/ansible-win-scoop](https://github.com/jamiemagee/ansible-win-scoop). Currently it can only install scoop, and install scoop packages, but I think it validates the concept. If there is interest I will continue development, with the intention of adding `win_scoop` as a core ansible module (Ideally feature complete with the chocolatey modules)

```yaml
- name: Install jq
  win_scoop:
    name: jq
```

See ansible/ansible#56084

JamieMagee:
##### SUMMARY
Initial implementation of a `win_scoop` module. Closes #50.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
win_scoop

##### ADDITIONAL INFORMATION
I've tested this manually, but I could use some information about how to write and structure unit + integration


JamieMagee:
@jborean93 can I request a review please?

plugins/modules/win_scoop.ps1
```diff
@@ -0,0 +1,230 @@
...
+# Copyright: (c) 2020, Jamie Magee <jamie.magee@gmail.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+#Requires -Module Ansible.ModuleUtils.ArgvParser
+#Requires -Module Ansible.ModuleUtils.CommandUtil
+#AnsibleRequires -CSharpUtil Ansible.Basic
+
+$spec = @{
+  options = @{
+    arch        = @{ type = "str"; choices = "32bit", "64bit" }
```
jborean93, line 12: It might be nice to have this as `architecture` or at least have an alias to that to match what win_chocolatey has.

plugins/modules/win_scoop.ps1
```diff
@@ -0,0 +1,230 @@
...
+$skip = $module.Params.skip
+$state = $module.Params.state
+
+$module.Result.rc = 0
+
+function Install-Scoop {
+
+  # Scoop doesn't have refreshenv like Chocolatey
+  # Let's try to update PATH first
+  $env:Path = [System.Environment]::GetEnvironmentVariable("Path", "Machine") + ";" + [System.Environment]::GetEnvironmentVariable("Path", "User")
```
jborean93, line 40: Why is this needed, by default PowerShell will have the Machine, User, and Process scoped env vars under the `env:` provider.

plugins/modules/win_scoop.ps1
```diff
@@ -0,0 +1,230 @@
...
+
+  return $scoop_app.Path
+}
+
+function Get-ScoopPackages {
+  param(
+    [Parameter(Mandatory = $true)] [string]$scoop_path
+  )
+
+  $command = Argv-ToString -arguments @("powershell.exe", $scoop_path, "export")
```
jborean93, line 103: It seems like scoop uses PowerShell scripts and is not an compiled binary. Shouldn't it be able to handle positional args as strings and we don't have to worry about escaping the arguments for the command line by just calling the script? For example this could be simplified with `$installed_packages = &$scoop_path export`.

Does scoop have a way of exporting actual objects and not strings, seems like an unfortunate omission if it isn't possible as it's bringing down the advantages that PowerShell offers.

plugins/modules/win_scoop.ps1
```diff
@@ -0,0 +1,230 @@
...
+  }
+
+  $installed_packages = $res.stdout -split "`n" |
+  Select-String '(.*?) \(v:(.*?)\) \[(.*?)\]' |
+  ForEach-Object {
+    New-Object PSObject -Property ([Ordered]@{
+        "Package" = $_.Matches[0].Groups[1].Value
+        "Version" = $_.Matches[0].Groups[2].Value
+        "Bucket"  = $_.Matches[0].Groups[3].Value
+      })
```
jborean93, line 120: Simpler, easier, and more PowerShell like to just do

```suggestion
    [PSCustomObject]@{
        Package = $_.Matches[0].Groups[1].Value
        Version = $_.Matches[0].Groups[2].Value
        Bucket  = $_.Matches[0].Groups[3].Value
    }
```

plugins/modules/win_scoop.ps1
```diff
@@ -0,0 +1,230 @@
...
+  if ($res.rc -ne 0) {
+    $module.Result.command = $command
+    $module.Result.rc = $res.rc
+    $module.Result.stdout = $res.stdout
+    $module.Result.stderr = $res.stderr
+    $module.FailJson("Error checking installed packages")
+  }
+
+  $installed_packages = $res.stdout -split "`n" |
+  Select-String '(.*?) \(v:(.*?)\) \[(.*?)\]' |
```
jborean93, line 114: This is a nit as it's your module but I find it's nice to at least indent the pipeline operations compared to the original call so we can see that this is running with pipeline input rather than a direct invocation of the cmdlet.

```powershell
$installed_packages = $res.stdout -split "`n" |
  Select-String '(.*?) \(v:(.*?)\) \[(.*?)\]' |
  ForEach-Object {
    ...
  }
```

plugins/modules/win_scoop.ps1
```diff
@@ -0,0 +1,230 @@
...
+  ForEach-Object {
+    New-Object PSObject -Property ([Ordered]@{
+        "Package" = $_.Matches[0].Groups[1].Value
+        "Version" = $_.Matches[0].Groups[2].Value
+        "Bucket"  = $_.Matches[0].Groups[3].Value
+      })
+  }
+
+  return $installed_packages
```
jborean93, line 123: Technically you can omit this line and just remove the `$installed_packages = ` part above. Anything output in that ForEach-Object will act the same here and you now no longer have a var that isn't really necessary.

plugins/modules/win_scoop.ps1
```diff
@@ -0,0 +1,230 @@
...
+}
+
+function Get-InstallScoopPackageArguments {
+  $arguments = [System.Collections.ArrayList]@()
```
jborean93, line 127: I know this is probably copied from win_chocolatey but if you use a generic list you don't need to discard the output from .Add()

```powershell
$arguments = [System.Collections.Generic.List[String]]@()

$arguments.Add('--arch')
$arguments.Add($arch)

...
```

In any case this comes back to an earlier point I made in a comment above. Because we are calling `scoop.ps1` we shouldn't have to escape these arguments. We are passing the string as an object using PowerShell's rules and I hope that scoop respects that.

plugins/modules/win_scoop.py
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2020, Jamie Magee <jamie.magee@gmail.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
```
jborean93, line 7: The `ANSIBLE_METADATA` section is no longer needed for modules and can be removed.

plugins/modules/win_scoop.py
```diff
@@ -0,0 +1,82 @@
...
+    description:
+    - Don't use the download cache
+    type: bool
+    default: no
+  purge:
+    description:
+    - Remove all persistent data
+    type: bool
+    default: no
+  skip:
```
jborean93, line 50: Skip is quite generic here, should be more specific like `skip_checksum` or `checksum_check` in case we need `skip` for some future purpose.

plugins/modules/win_scoop.py
```diff
@@ -0,0 +1,82 @@
...
+  state:
+    description:
+    - State of the package on the system.
+    - When C(absent), will ensure the package is not installed.
+    - When C(present), will ensure the package is installed.
+    type: str
+    choices: [ absent, present ]
+    default: present
+seealso:
+- module: win_chocolatey
```
jborean93, line 64: The `win_chocolatey` module is no longer part of this collection so we should use the fully qualified collection name

```suggestion
- module: chocolatey.chocolatey.win_chocolatey
```

plugins/modules/win_scoop.py
```diff
@@ -0,0 +1,82 @@
...
+- name: Scoop main bucket
+  description: GitHub repository for the main bucket
+  link: https://github.com/ScoopInstaller/Main
+author:
+- Jamie Magee (@JamieMagee)
+'''
+
+EXAMPLES = r'''
+- name: Install jq.
+  win_scoop:
```
jborean93, line 80: We are trying to start using the fully qualified collection name in the example so this would become

```suggestion
  community.windows.win_scoop:
```

plugins/modules/win_scoop.ps1
JamieMagee: I was torn between `arch` and `architecture` here. `arch` is what scoop itself uses. From the help text:

> -a, --arch <32bit|64bit>  Use the specified architecture, if the app supports it

But I will change it if you think it's better to have a common parameter name between the different package managers.

plugins/modules/win_scoop.ps1
JamieMagee: I'd be fine with this change on it's own, but with the below change about omitting the explicit return value, I think it looks better without over-indenting the first line.

plugins/modules/win_scoop.ps1
JamieMagee: It's a bit of a hack. Scoop doesn't update the current session with environment variable changes. Without this line, or rather without the change on line 79, scoop isn't found in the path. So even after scoop is installed here, it's not found for any of the subsequent commands.

I know this isn't the best, but I am open to alternatives

plugins/modules/win_scoop.ps1
JamieMagee: I gave this a go, and `scoop export` is working fine. However I am struggling with the `install` and `uninstall` commands. Here's a snipped from `Install-ScoopPackage`

```pwsh
$arguments = [System.Collections.Generic.List[String]]@("install")
$arguments.AddRange($packages)

$common_args = Get-InstallScoopPackageArguments
$arguments.AddRange($common_args)

$stdout = & $scoop_path @arguments
```

Using this I'm getting some weird errors from scoop: `Can't shim 'bin\\basename.exe': File doesn't exist.` Can you see something wrong with my PowerShell that is obviously wrong? If not, I think I'll stick with the command escaping for now.

As for PowerShell objects scoop does have this, but it's undocumented, the parameters are binary flags instead of the parameters the CLI expects, and it's recommended not to directly use the scoop internals.

plugins/modules/win_scoop.ps1
jborean93: You can have both, `@{ type = 'str'; choices = '32bit', '64bit'; aliases = 'arch' }`. I would just recommend the full module option matches the "convention" we already have with win_chocolatey.

plugins/modules/win_scoop.ps1
jborean93: I suppose this is simpler then just trying to add the scoop path to `$env:PATH` as it will get whatever changes the scoop installer adds.

plugins/modules/win_scoop.ps1
jborean93: The trouble that I find with this form is that it makes it look like `Select-String` and `ForEach-Object` their own call without any input going into it. I had to double take why the `ForEach-Object` was working without the `($var in $coll)` part but then realised it was the `ForEach-Object` cmdlet instead.

plugins/modules/win_scoop.py
jborean93: In this case, I think `skip` is just a bit too vague. To me it pretty much says skip this package which could be conditional for any reason. While modules should closely reflect the underlying resource it manages we should try and ensure we keep some consistency with existing Ansible conventions as well as paper over some of the more confusing aspects of a resource. In this case `skip_checksum` tells the Ansible author what exactly it is skipping whereas seeing `skip` in a playbook is only really obvious to people who have used scoop before.

plugins/modules/win_scoop.ps1
jborean93: > Can you see something wrong with my PowerShell that is obviously wrong?

I can't see anything wrong with that snippet, it definitely feel like something is up with the environment PATHs.

> If not, I think I'll stick with the command escaping for now.

No worries, thought it was worth a shot to try and reduce some complexity but if it doesn't work then let's ignore it.

> As for PowerShell objects scoop does have this, but it's undocumented, the parameters are binary flags instead of the parameters the CLI expects, and it's recommended not to directly use the scoop internals.

That's a pity but it is what it is, thanks for letting me know.

JamieMagee: Okay. Thanks for your input. I'll resolve this comment.

plugins/modules/win_scoop.ps1
JamieMagee: I'm resolving this for now, but open to any alternatives.

plugins/modules/win_scoop.ps1
```diff
@@ -0,0 +1,225 @@
...
+  # Packages can be in a broken state where they don't appear scoop export
+  Uninstall-ScoopPackage -scoop_path $scoop_path -packages $name
+}
+
+if ($state -in @("present")) {
+  $missing_packages = [System.Collections.ArrayList]@()
+  foreach ($package in $name) {
+    if ($installed_packages.Package -notcontains $package) {
+      $missing_packages.Add($package)
+    }
```
jborean93, line 217: You are still using an ArrayList here so `.Add()` will be producing unwanted output. While things may work it can break in certain situations where the output is mixed with other data. You can really simplify this by avoiding the list altogether

```suggestion
  $missing_packages = foreach ($package in $name) {
    if ($installed_packages.Package -notcontains $packages) {
      $package
    }
  }
```

Anything in the output in the foreach loop is now assigned to the var `$missing_packages`.

plugins/modules/win_scoop.ps1
```diff
@@ -0,0 +1,225 @@
...
+  options = @{
+    architecture  = @{ type = "str"; choices = "32bit", "64bit"; aliases = @(, "arch") }
+    independent   = @{type = "bool"; default = $false }
+    global        = @{ type = "bool"; default = $false }
+    name          = @{ type = "list"; elements = "str"; required = $true }
+    no_cache      = @{type = "bool"; default = $false }
+    purge         = @{type = "bool"; default = $false }
+    skip_checksum = @{type = "bool"; default = $false }
+    state         = @{ type = "str"; default = "present"; choices = "present", "absent" }
+  }
```
jborean93, line 20: So I see you are using `$module.CheckMode` checks when installing scoop but a module by default does not support check mode. You need to add `supported_check_mode = $true` to `$spec`. When you add it you also need to make sure that the install/uninstall doesn't actually run in check mode, or add a `-whatif` like flag to scoop if it supports it.

JamieMagee:
@jborean93 Thanks for the thorough review. This is my first custom ansible module, and I appreciate the guidance.

> Thanks for the integration tests, they are extremely helpful. Do you know how hard it is to create your own offline packages? It would be nice to reduce our reliance on an internet connection as much as possible.

I realised I've missed this earlier. It is possible to create custom packages, but after looking at it for a bit, the `win_scoop_bucket` module in #94 is going to require at a minimum `git` and `7zip` to be installed by `win_scoop`. Either way, there's going to be some reliance on an internet connection , but would you like me to switch the test packages for `win_scoop` to be `git` and `7zip` instead of `grep` and `sed` as it is now?

jborean93:
No worries, glad to have you working on all this. Because we can't remove our reliance on a working internet connection entirely we typically try to reduce the number of external calls we make to have the tests be as stable as they can. We've had issues in the past where external services go down which bring down our tests and they are never fun to deal with.

In saying that we should probably just continue ahead with using the current way you've written the tests and only create our own packages if that proves to be too unstable in the future.

One of the other benefits of using your own packages is you can control the install/uninstall behaviour for further checks and validation. This was quite useful when it came to testing `win_chocolatey` to test things like versions, extra args used, etc but for scoop there doesn't seem to be a complex bunch of options like chocolatey has. I think for where this module is at we really don't need something like that, just something to think about in the future.

plugins/modules/win_scoop.ps1
```diff
@@ -0,0 +1,234 @@
...
+  $arguments.AddRange($packages)
+
+  $common_args = Get-InstallScoopPackageArguments
+  $arguments.AddRange($common_args)
+
+  $command = Argv-ToString -arguments $arguments
+
+  if (-not $module.CheckMode) {
+    $res = Run-Command -Command $command
+    $module.Result.rc = $res.rc
```
jborean93, line 160: So I see we aren't checking the return code, is there any error validation that we can do to make sure the install worked?

plugins/modules/win_scoop.ps1
```diff
@@ -0,0 +1,234 @@
...
+
+    if ($module.Verbosity -gt 1) {
+      $module.Result.stdout = $res.stdout
+    }
+    if (-not ($res.stdout -match "ERROR '(.*?)' isn't installed.")) {
```
jborean93, line 203: Will this be a problem with other locales?

plugins/modules/win_scoop.ps1
```diff
@@ -0,0 +1,234 @@
...
+
+if ($state -in @("absent")) {
+  # Always attempt uninstall
+  # Packages can be in a broken state where they don't appear scoop export
+  Uninstall-ScoopPackage -scoop_path $scoop_path -packages $name
+}
+
+if ($state -in @("present")) {
+  $missing_packages = foreach ($package in $name) {
+    if ($installed_packages.Package -notcontains $packages) {
```
jborean93, line 224: Looks like this is why the tests aren't passing, sorry for the faulty suggestion in the last round of commits.

```suggestion
    if ($installed_packages.Package -notcontains $package) {
```

plugins/modules/win_scoop.ps1
JamieMagee: Scoop isn't localised. See [here](https://github.com/lukesampson/scoop/blob/08af9ff6e7f7b017701c0c9114294b13e1e83fb8/lib/core.ps1#L670-L672) for this specific error message.

plugins/modules/win_scoop.ps1
JamieMagee: I copied the return code check from `Get-ScoopPackages` to `Install-ScoopPackage` and `Uninstall-ScoopPackage`.

jborean93:
Thanks for working on this @JamieMagee.

JamieMagee:
You're welcome. Thanks for all the help!

When are you planning to cut a new release?

jborean93:
It will be sometime before the 2.10 release of Ansible so within a month hopefully

PR #93, opened 2020-06-02
win_domain_computer - add offline domain join support

briantist:
##### SUMMARY

Looking to add support for offline domain join, similar to the `ADComputer` DSC resource.

Offline Domain Join is usually done with `djoin.exe` and lets you pre-create a computer and get back an opaque BLOB, which you can use on a target machine or machine image to join the machine to the domain, without having to contact a domain controller.

[A bug I recently found in the DSC resource](https://github.com/dsccommunity/ActiveDirectoryDsc/issues/608) led me to look at adding the support directly to Ansible instead of continuing to use the DSC resource.

Main changes:
- Add support for doing an offline domain join provision
- Added tests for ODJ
- Optionally return the BLOB from the module

Other changes:
- Leaving off the `$` from `sam_account_name` is no longer an error; it will be added for you
- fixed up tests which were completely broken and unusable, updated their comments to better inform how they could be used locally (they don't run in CI)
- fixed up some syntax and PowerShell calls
- touched up documentation, added examples and notes

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

win_domain_computer

### DIFF
--- a/plugins/modules/win_domain_computer.ps1
+++ b/plugins/modules/win_domain_computer.ps1
@@ -206,6 +206,10 @@ Function Invoke-OfflineDomainJoin {
       $result.odj_blob_path = $BlobPath.FullName
     }
 
+    if (-not $BlobPath.Directory.Exists) {
+      Fail-Json -obj $result -message "BLOB path directory '$($BlobPath.Directory.FullName)' doesn't exist."
+    }
+
     if ($PSCmdlet.ShouldProcess($argstring)) {
       try {
         $djoin_result = Run-Command -command $invocation
@@ -224,7 +228,7 @@ Function Invoke-OfflineDomainJoin {
         }
       }
       finally {
-        if ($output) {
+        if ($output -and $BlobPath.Exists) {
           $BlobPath.Delete()
         }
       }
--- a/plugins/modules/win_domain_computer.ps1
+++ b/plugins/modules/win_domain_computer.ps1
@@ -20,9 +20,9 @@ $check_mode = Get-AnsibleParam -obj $params -name "_ansible_check_mode" -type "b
 $diff_support = Get-AnsibleParam -obj $params -name "_ansible_diff" -type "bool" -default $false
 
 $name = Get-AnsibleParam -obj $params -name "name" -failifempty $true -resultobj $result
-$sam_account_name = Get-AnsibleParam -obj $params -name "sam_account_name" -default "$name$"
+$sam_account_name = Get-AnsibleParam -obj $params -name "sam_account_name" -default "${name}$"
 If (-not $sam_account_name.EndsWith("$")) {
-  Fail-Json -obj $result -message "sam_account_name must end in $"
+  $sam_account_name = "${sam_account_name}$"
 }
 $enabled = Get-AnsibleParam -obj $params -name "enabled" -type "bool" -default $true
 $description = Get-AnsibleParam -obj $params -name "description" -default $null
@@ -74,12 +74,14 @@ Function Get-InitialState($desired_state) {
       @extra_args
   } Catch { $null }
   If ($computer) {
+      $null,$current_ou = $computer.DistinguishedName -split '(?<=[^\\](?:\\\\)*),'
+      $current_ou = $current_ou -join ','
+
       $initial_state = [ordered]@{
         name = $computer.Name
         sam_account_name = $computer.SamAccountName
         dns_hostname = $computer.DNSHostName
-        # Get OU from regexp that removes all characters to the first ","
-        ou = $computer.DistinguishedName -creplace "^[^,]*,",""
+        ou = $current_ou
         distinguished_name = $computer.DistinguishedName
         description = $computer.Description
         enabled = $computer.Enabled
@@ -163,7 +165,7 @@ Function Remove-ConstructedState($initial_state) {
 }
 
 # ------------------------------------------------------------------------------
-Function are_hashtables_equal($x, $y) {
+Function Test-HashtableEquality($x, $y) {
   # Compare not nested HashTables
   Foreach ($key in $x.Keys) {
       If (($y.Keys -notcontains $key) -or ($x[$key] -cne $y[$key])) {
@@ -183,17 +185,17 @@ $initial_state = Get-InitialState($desired_state)
 
 If ($desired_state.state -eq "present") {
     If ($initial_state.state -eq "present") {
-      $in_desired_state = are_hashtables_equal $initial_state $desired_state
+      $in_desired_state = Test-HashtableEquality -X $initial_state -Y $desired_state
 
       If (-not $in_desired_state) {
-        Set-ConstructedState $initial_state $desired_state
+        Set-ConstructedState -initial_state $initial_state -desired_state $desired_state
       }
     } Else { # $desired_state.state = "Present" & $initial_state.state = "Absent"
-      Add-ConstructedState($desired_state)
+      Add-ConstructedState -desired_state $desired_state
     }
   } Else { # $desired_state.state = "Absent"
     If ($initial_state.state -eq "present") {
-      Remove-ConstructedState($initial_state)
+      Remove-ConstructedState -initial_state $initial_state
     }
   }
 
--- a/plugins/modules/win_domain_computer.ps1
+++ b/plugins/modules/win_domain_computer.ps1
@@ -21,6 +21,7 @@ $params = Parse-Args $args -supports_check_mode $true
 
 $check_mode = Get-AnsibleParam -obj $params -name "_ansible_check_mode" -type "bool"  -default $false
 $diff_support = Get-AnsibleParam -obj $params -name "_ansible_diff" -type "bool" -default $false
+$temp = Get-AnsibleParam -obj $params -name '_ansible_remote_tmp' -type 'path' -default $env:TEMP
 
 $name = Get-AnsibleParam -obj $params -name "name" -failifempty $true -resultobj $result
 $sam_account_name = Get-AnsibleParam -obj $params -name "sam_account_name" -default "${name}$"
@@ -34,11 +35,9 @@ $domain_password = Get-AnsibleParam -obj $params -name "domain_password" -type "
 $domain_server = Get-AnsibleParam -obj $params -name "domain_server" -type "str"
 $state = Get-AnsibleParam -obj $params -name "state" -ValidateSet "present","absent" -default "present"
 
-$do_odj = Get-AnsibleParam -obj $params -name "offline_domain_join" -type "bool" -default $false
-if ($do_odj) {
-  $odj_return_blob = Get-AnsibleParam -obj $params -name "odj_return_blob" -type "bool" -default $false
-  $odj_blob_path = Get-AnsibleParam -obj $params -name "odj_blob_path" -type "str" -failifempty (-not $odj_return_blob) -default $null
-}
+$odj_action = Get-AnsibleParam -obj $params -name "offline_domain_join" -type "str" -ValidateSet "none","output","path" -default "none"
+$_default_blob_path = Join-Path -Path $temp -ChildPath ([System.IO.Path]::GetRandomFileName())
+$odj_blob_path = Get-AnsibleParam -obj $params -name "odj_blob_path" -type "str" -default $_default_blob_path
 
 $extra_args = @{}
 if ($null -ne $domain_username) {
@@ -164,9 +163,10 @@ Function Invoke-OfflineDomainJoin {
     [System.Collections.IDictionary]
     $desired_state ,
 
-    [Parameter()]
-    [bool]
-    $ReturnBlob = $false ,
+    [Parameter(Mandatory=$true)]
+    [ValidateSet('none','output','path')]
+    [String]
+    $Action ,
 
     [Parameter()]
     [System.IO.FileInfo]
@@ -174,14 +174,16 @@ Function Invoke-OfflineDomainJoin {
   )
 
   End {
-    $dns_domain = $desired_state.dns_hostname -replace '^[^.]+\.'
-
-    $delete = -not $BlobPath
-    if ($delete) {
-      $BlobPath = [System.IO.Path]::GetTempFileName()
+    if ($Action -eq 'none') {
+      return
     }
 
+    $dns_domain = $desired_state.dns_hostname -replace '^[^.]+\.'
+
+    $output = $Action -eq 'output'
+
     $arguments = @(
+      'djoin.exe'
       '/PROVISION'
       '/REUSE'  # we're pre-creating the machine normally to set other fields, then overwriting it with this
       '/DOMAIN'
@@ -194,31 +196,37 @@ Function Invoke-OfflineDomainJoin {
       $BlobPath.FullName
     )
 
-    $argstring = Argv-ToString -arguments $arguments
-    $exe = 'djoin.exe'
-    $invocation = "$exe $argstring"
+    $invocation = Argv-ToString -arguments $arguments
     $result.djoin = @{
       invocation = $invocation
     }
-    $result.odj_blob = [String]::Empty
+    $result.odj_blob = ''
 
-    if ($PSCmdlet.ShouldProcess($exe, $argstring)) {
-      $djoin_result = Run-Command -command $invocation
-      $result.djoin.rc = $djoin_result.rc
-      $result.djoin.stdout = $djoin_result.stdout
-      $result.djoin.stderr = $djoin_result.stderr
+    if ($Action -eq 'path') {
+      $result.odj_blob_path = $BlobPath.FullName
+    }
 
-      if ($djoin_result.rc) {
-        Fail-Json -obj $result -message "Problem running djoin.exe. See returned values."
+    if ($PSCmdlet.ShouldProcess($argstring)) {
+      try {
+        $djoin_result = Run-Command -command $invocation
+        $result.djoin.rc = $djoin_result.rc
+        $result.djoin.stdout = $djoin_result.stdout
+        $result.djoin.stderr = $djoin_result.stderr
+
+        if ($djoin_result.rc) {
+          Fail-Json -obj $result -message "Problem running djoin.exe. See returned values."
+        }
+
+        if ($output) {
+          $bytes = [System.IO.File]::ReadAllBytes($BlobPath.FullName)
+          $data = [Convert]::ToBase64String($bytes)
+          $result.odj_blob = $data
+        }
       }
-
-      if ($ReturnBlob) {
-        $data = Get-Content -LiteralPath $BlobPath.FullName -Encoding Unicode -Raw
-        $result.odj_blob = $data.TrimEnd("`0")
-      }
-
-      if ($delete) {
-        $BlobPath.Delete()
+      finally {
+        if ($output) {
+          $BlobPath.Delete()
+        }
       }
     }
   }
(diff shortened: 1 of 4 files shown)


plugins/modules/win_domain_computer.ps1
```diff
@@ -20,9 +23,9 @@ $check_mode = Get-AnsibleParam -obj $params -name "_ansible_check_mode" -type "b
 $diff_support = Get-AnsibleParam -obj $params -name "_ansible_diff" -type "bool" -default $false
 
 $name = Get-AnsibleParam -obj $params -name "name" -failifempty $true -resultobj $result
-$sam_account_name = Get-AnsibleParam -obj $params -name "sam_account_name" -default "$name$"
+$sam_account_name = Get-AnsibleParam -obj $params -name "sam_account_name" -default "${name}$"
```
jborean93, line 26: Curious is the `${name}` change just a style change or does this actually change the behaviour. I would have thought the original code was fine?

plugins/modules/win_domain_computer.py
```diff
@@ -47,6 +49,7 @@
     description:
       - Specifies the X.500 path of the Organizational Unit (OU) or container
         where the new object is created. Required when I(state=present).
+      - Special characters must be escaped, see U(https://docs.microsoft.com/en-us/previous-versions/windows/desktop/ldap/distinguished-names)
```
jborean93, line 52: Is this just a limitation of the module or special characters in OU's in general?

Also you can use `L(Link name,URL)` to shorten the URL as it appears in the docs

plugins/modules/win_domain_computer.py
```diff
@@ -86,6 +89,28 @@
     type: str
     choices: [ absent, present ]
     default: present
+  offline_domain_join:
+    description:
+      - Provisions a computer in the directory and provides a BLOB file that can be used on the target computer/image to join it to the domain while offline.
+      - If the computer already exists, no BLOB will be created/returned, and the module will operate as it would have without offline domain join.
+    type: bool
+  odj_return_blob:
```
jborean93, line 97: Instead of having 2 options for this we could combine `offline_domain_join` and `obj_return_blob` into 1 option like `offline_join_blob`. This option will be a string that accepts 3 options

* `none` - (default) no blob is generated/returned
* `output` - Outputs the blob in the module return json
* `path` - Outputs the blob in the path specified by `obj_blob_path`

I'm in no way set on the names here, just suggestions to get the conversation going and to show what I mean.

plugins/modules/win_domain_computer.py
```diff
@@ -86,6 +89,28 @@
...
+      - The BLOB file contains a terminating null character which will not be included. See Notes.
+  odj_blob_path:
+    description:
+      - The path to the file where the BLOB will be saved.
+      - Required when I(offline_domain_join=True) and I(odj_return_blob=False).
+      - If using I(odj_return_blob=True) this can be omitted, and a temporary file will be used and then destroyed.
+      - If I(odj_return_blob=True) and this is specified, it will be used and file will remain.
+notes:
+  - "For more information on Offline Domain Join
+    see U(https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2008-R2-and-2008/dd392267(v=ws.10))"
```
jborean93, line 110: You will probably have to remove the `(v=ws.10)` part of the URL. There's no way I know off to escape `(` and `)` in a URL for `U()`. Even better you can use `L(Link text,URL)` to shorten the URL as it is shown in the module docs.

plugins/modules/win_domain_computer.py
```diff
@@ -86,6 +89,28 @@
...
+  - When using the ODJ BLOB to join a computer to the domain, it must be written out to a file.
```
jborean93, line 111: Is this a limitation of `djoin.exe`? Is it a problem when you PInvoke?

plugins/modules/win_domain_computer.ps1
```diff
@@ -146,6 +157,73 @@ Function Add-ConstructedState($desired_state) {
...
+Function Invoke-OfflineDomainJoin {
+  [CmdletBinding(SupportsShouldProcess=$true)]
+  param(
+    [Parameter(Mandatory=$true)]
+    [System.Collections.IDictionary]
+    $desired_state ,
+
+    [Parameter()]
+    [bool]
+    $ReturnBlob = $false ,
```
jborean93, line 169: Any reason why this can't be a `-Switch` to be more powershell-like?

plugins/modules/win_domain_computer.ps1
```diff
@@ -146,6 +157,73 @@ Function Add-ConstructedState($desired_state) {
...
 }
 
+Function Invoke-OfflineDomainJoin {
+  [CmdletBinding(SupportsShouldProcess=$true)]
+  param(
+    [Parameter(Mandatory=$true)]
+    [System.Collections.IDictionary]
+    $desired_state ,
+
+    [Parameter()]
```
jborean93, line 167: Minor nit, `[Parameter()]` isn't really needed as that's the default for an advanced function.

plugins/modules/win_domain_computer.ps1
```diff
@@ -146,6 +157,73 @@ Function Add-ConstructedState($desired_state) {
...
+    [bool]
+    $ReturnBlob = $false ,
+
+    [Parameter()]
+    [System.IO.FileInfo]
+    $BlobPath
+  )
+
+  End {
```
jborean93, line 176: If you don't have a being/process/end block just omit it entirely. The default in the absence of these blocks is `End {}`

plugins/modules/win_domain_computer.ps1
```diff
@@ -146,6 +157,73 @@ Function Add-ConstructedState($desired_state) {
...
+    $dns_domain = $desired_state.dns_hostname -replace '^[^.]+\.'
+
+    $delete = -not $BlobPath
+    if ($delete) {
+      $BlobPath = [System.IO.Path]::GetTempFileName()
```
jborean93, line 181: While it's not as easy to access in the older style modules you can access the temp path set by Ansible with

```powershell
$temp = Get-AnsibleParam -obj $params -name '_ansible_remote_tmp' -type 'path' -default $env:TEMP
```

From there you can do `$BlobPath = Join-Path -Path $temp -ChildPath ([IO.Path]::GetRandomFileName())`. The only other thing I suggest is to wrap this in a try/catch to ensure `$BlobPath` is deleted if it exists and we don't need to keep it

plugins/modules/win_domain_computer.ps1
```diff
@@ -146,6 +157,73 @@ Function Add-ConstructedState($desired_state) {
...
+      $desired_state.sam_account_name.TrimEnd('$')  # this machine name is the short name
+      '/MACHINEOU'
+      $desired_state.ou
+      '/SAVEFILE'
+      $BlobPath.FullName
+    )
+
+    $argstring = Argv-ToString -arguments $arguments
+    $exe = 'djoin.exe'
+    $invocation = "$exe $argstring"
```
jborean93, line 199: You may as well put `djoin.exe` as the first value in `$arguments`, that way you can just do `$invocation = Argv-ToString -arguments $arguments`

plugins/modules/win_domain_computer.ps1
```diff
@@ -146,6 +157,73 @@ Function Add-ConstructedState($desired_state) {
...
+    $result.djoin = @{
+      invocation = $invocation
+    }
+    $result.odj_blob = [String]::Empty
```
jborean93, line 203: Why not `= ''`?

plugins/modules/win_domain_computer.ps1
```diff
@@ -146,6 +157,73 @@ Function Add-ConstructedState($desired_state) {
...
+      $result.djoin.rc = $djoin_result.rc
+      $result.djoin.stdout = $djoin_result.stdout
+      $result.djoin.stderr = $djoin_result.stderr
+
+      if ($djoin_result.rc) {
+        Fail-Json -obj $result -message "Problem running djoin.exe. See returned values."
+      }
+
+      if ($ReturnBlob) {
+        $data = Get-Content -LiteralPath $BlobPath.FullName -Encoding Unicode -Raw
```
jborean93, line 216: What's the actual format of the contents? Just wondering whether we should return the raw bytes as a base64 string instead of trying to return the string contents and removing bytes to achieve that.

plugins/modules/win_domain_computer.ps1
briantist: Style only; a little clearer imo. I can revert but in this case I don't think it matters much git blame wise because of the collection split.

jborean93: That's ok, it's a benign change and I was just curious if it was done for a particular reason.

plugins/modules/win_domain_computer.py
briantist: This is a general thing you'd have to do when passing OU names to modules/libraries. It's not really specific to this one, but thought it was worth pointing out.

I thought the `L()` syntax only worked on links within docs? I had always wanted to use it for outside links but thought it was impossible:

https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#linking-within-module-documentation
> For modules in a collection, you can only use L() and M() for content within that collection. Use U() to refer to content in a different collection.

plugins/modules/win_domain_computer.py
briantist: Will that cause a problem? I can't remove it completely (already tried, 404). I could try URL escape sequences, or I could use GitHub's URL shortener if that's acceptable.

See  my other comment about `L()`, could use some clarification there.

plugins/modules/win_domain_computer.py
briantist: Do you mean having it written to a file? That's a limitation of `djoin.exe`. With p/invoke you'd be giving it a buffer directly (probably the base64-decoded contents of what we get from `djoin.exe`); but that would happen outside this module anyway.

plugins/modules/win_domain_computer.ps1
briantist: No reason; I can change that.

plugins/modules/win_domain_computer.ps1
briantist: I'm aware; I prefer it personally for consistency, but I don't feel strongly enough about it to keep it if you prefer it omitted.

plugins/modules/win_domain_computer.ps1
briantist: Like `[Parameter()]` it's my own style to make it explicit, but I can remove if you prefer.

plugins/modules/win_domain_computer.ps1
briantist: The only reason I ended up keeping it separate was for `ShouldProcess()` but I can use the single argument form of that and just put it together.

plugins/modules/win_domain_computer.ps1
briantist: No particular reason; I can change it

plugins/modules/win_domain_computer.ps1
briantist: I was actually talking about this with a co-worker today. The contents are themselves a base64 encoded string, except that the file ends with a null character (which seems to cause issues trying to send that through Ansible raw). 

I could re-encode the entire file with base64 and return it, but then in order to use it, on the controller, it would need to be decoded and the null character may cause issues again.

So in considering this, I think it ends up being easier to use it this way (I added an example in the docs), but happy to consider alternatives.

plugins/modules/win_domain_computer.py
jborean93: Links within the docs are meant to use the new-ish R() syntax https://github.com/ansible/ansible/pull/69832. I'll double check but I've used `L()` before as an equivalent for `[label](link)`.

plugins/modules/win_domain_computer.ps1
jborean93: No worries, just thought I would mention it.

plugins/modules/win_domain_computer.ps1
jborean93: Oh I didn't realise that, still the executable is always `djoin.exe` so there's no reason why you can't just hardcode that in ShouldProcess().

plugins/modules/win_domain_computer.ps1
jborean93: In this case I think you should, not that it's wrong what you have it's just simpler to see that's an empty string than the .NET static field.

plugins/modules/win_domain_computer.ps1
jborean93: I think in this case you should, boolean parameters aren't very common in PowerShell and when they are used they are typically can just be a -Switch.

plugins/modules/win_domain_computer.py
jborean93: Yep, was just curious if having it written to a file was a limitation of what djoin.exe offers or whether the underlying Win32 calls also has the restrictions.

plugins/modules/win_domain_computer.py
jborean93: > I can't remove it completely (already tried, 404)

That's weird, definitely works for other Microsoft docs but obviously not this one. You might have to use a URL shortened if you can't URL escape the brackets

plugins/modules/win_domain_computer.py
briantist: Not completely opposed, but it would remove the possibility of specifying a file to keep and also returning the blob. But I doubt that combination is very useful so this would probably be fine.

briantist:
> Just for my general understanding the offline stuff is still creating the domain object in AD, it just handles all the logic to generate the offline join blob right?
Yes that's correct.

plugins/modules/win_domain_computer.ps1
jborean93: Sigh what a crappy UX for djoin but it's not like we can fix that. Personally I think we should still return the base64 encoded bytes of the file and not the massaged contents. Here are the disadvantages I see with the current process:

* The actual file is UTF-16 encoded, the user of this function would have to know to use Unicode encoding of the value when creating the file on the host to join
* The `\u0000` value on the end cannot be transferred across which means it's more manual work by the end user to rebuild the file
* This is scraping the barrel but it is also less susceptible to changes made to the format in the future. Personally I doubt MS would change it but if they do add more binary data it would be a problem here

Whereas returning the base64 string has the advantages

* Is a pure representation of the file bytes, user's know need to care about the encoding unless they want to convert to a string themselves
* Single step in PowerShell to create a file from it
* Not reliant on the format staying the same

While some of those points are sort off solved by the examples you've it's still extra knowledge they need to be aware of. If it was just that the file needs to be UTF-16 encoded then I would be fine with returning the data as is, but because it needs both the null char and UTF-16 encoding I would err on the side of base64'ing the file itself.

In this case the example in the docs would be

```yaml
- name: Join a workgroup computer to the domain
  vars:
    target_blob_file: 'C:\ODJ\blob.txt'
  ansible.windows.win_shell: |
    $blob = [Convert]::FromBase64String('{{ computer_status.obj_blob }}')
    [IO.File]::WriteAllBytes('{{ target_blob_file }}', $blob)
    & djoin.exe --% /RequestODJ /LoadFile '{{ target_blob_file }}' /LocalOS /WindowsPath "%SystemRoot%"
```

The only missing part of all this was if copy/win_copy supported setting the content has base64 data and it did all that for you. Unfortunately that won't work today and you would have to create the file on the controller manually defeating and then using win_copy making things more complicated.

plugins/modules/win_domain_computer.py
briantist: Oh I see on output? That's also a limitation of `djoin.exe` (sort of). It has an option to write the content to stdout, but you still have to give it a file to write to anyway, and the stdout has other stuff with it, so it's not that useful for programmatic use. The API doesn't deal with files at all, so on the provisioning part the API would have given the data back in a buffer too.

jborean93: No worries, I was curious because I've never used this method before.

plugins/modules/win_domain_computer.ps1
briantist: Ok yeah, that solves some of what I saw as a disadvantage of re-base64 encoding, is that even preserving the utf-16 representation, the consumer would still have to know upon doing a decode to also ensure they use utf-16. But side-stepping putting it in a string by just using the raw bytes of the decode makes that more appealing; I'll go with that.

I would consider adding the support to `win_domain_membership` but in my actual use case I don't use that module so it's tough for me to find time, especially to do the testing needed for the implementation.

I actually use this blob to perform a join on deploy of an AMI in AWS; putting a similar PowerShell snippet as above into user data (we never have to retrieve the generated admin password or do any further config before the instance joins itself to the domain and we can connect with domain creds).

I could at least write up a feature request that could serve as a guide in case someone else is interested in taking it up.

jborean93: > I would consider adding the support to win_domain_membership but in my actual use case I don't use that module so it's tough for me to find time, especially to do the testing needed for the implementation.

Totally understandable, just contributing this stuff is excellent and adds value to the module. Please don't feel pressured to add further features to `win_domain_membership`.

> I could at least write up a feature request that could serve as a guide in case someone else is interested in taking it up.

You can, my only concern is that this will just become another feature request that won't be worked on because we have similar time constraints. Definitely won't stop you from doing it though, just don't feel obliged to create it.

plugins/modules/win_domain_computer.py
briantist: Went with this, with the slight modification of `odj_blob_path` always being optional and using a temp file if not specified. In `output` mode if a file is specified it will use that, but this mode always deletes the created file. In `path` mode it always leaves the file (even if it's a temp file).

plugins/modules/win_domain_computer.py
briantist: URL encodng the parentheses seemed to work (also switched to `L()`).

plugins/modules/win_domain_computer.ps1
briantist: No longer needed as there's no bool/switch param.

plugins/modules/win_domain_computer.ps1
briantist: This all worked fine; the content is of course a lot larger now but otherwise looks good.

plugins/modules/win_domain_computer.ps1
```diff
@@ -146,6 +156,82 @@ Function Add-ConstructedState($desired_state) {
...
+
+        if ($output) {
+          $bytes = [System.IO.File]::ReadAllBytes($BlobPath.FullName)
+          $data = [Convert]::ToBase64String($bytes)
+          $result.odj_blob = $data
+        }
+      }
+      finally {
+        if ($output) {
+          $BlobPath.Delete()
```
jborean93, line 228: It seems like this file may not actually exist in a failure scenario. When looking at the code we are passing through a path which could be defined by the end user. If `djoin.exe` hasn't created the file for some reason (say it failed) then this will fail as well hiding the actual error.

Also, I'm unsure whether we need to check if the parent directory exists and fail if it doesn't. Does `djoin.exe` create any intermediate parent dirs that don't exist?

briantist: Good call; `djoin` will not create intermediate paths. I've added a check for that.

One of the things I like about the `FileInfo.Delete()` method is that it doesn't fail if the file doesn't exist; but to your point it will fail if the parent directory doesn't exist; so added a check for file existence anyway.

Also updated the docs to indicate the parent of the BLOB path needs to exist.

PR #111, opened 2020-07-07
Add checks for globally installed packages

JamieMagee:
##### SUMMARY
The main fix here is to check if packages are installed globally, as well as the architecture of the installed package.

Scoop allows packages to be installed locally and globally simultaneously, but not different architectures installed locally or different architectures installed globally e.g. you can have dotnet-sdk 32bit locally, and dotnet-sdk 64bit globally.

Also a couple of minor fixes around spelling.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_scoop

### DIFF
--- a/changelogs/changelog.yaml
+++ b/changelogs/changelog.yaml
@@ -26,6 +26,7 @@ releases:
       - win_iis_website - Actually restart the site when ``state=restarted`` - https://github.com/ansible/ansible/issues/63828
       - win_partition - Fix invalid variable name causing a failure on checks - https://github.com/ansible/ansible/issues/62401
       - win_partition - don't resize partitions if size difference is < 1 MiB
+      - win_scoop - add checks for globally installed packages for better idempotency checks
       - win_timezone - Allow for _dstoff timezones
       - win_unzip - Fix support for paths with square brackets not being detected
         properly
--- a/changelogs/changelog.yaml
+++ b/changelogs/changelog.yaml
@@ -26,7 +26,6 @@ releases:
       - win_iis_website - Actually restart the site when ``state=restarted`` - https://github.com/ansible/ansible/issues/63828
       - win_partition - Fix invalid variable name causing a failure on checks - https://github.com/ansible/ansible/issues/62401
       - win_partition - don't resize partitions if size difference is < 1 MiB
-      - win_scoop - add checks for globally installed packages for better idempotency checks
       - win_timezone - Allow for _dstoff timezones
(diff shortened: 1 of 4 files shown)


jborean93:
My guess the test here is failing because the result of `Get-ScoopPackages` only returns 1 package hence `$installed_packages` is just a `PSCustomObject. If you change https://github.com/ansible-collections/community.windows/blob/480bf7496110b0f10a4c80cdc8d1054b9340be94/plugins/modules/win_scoop.ps1#L228 to

```powershell
$installed_packages = @(Get-ScoopPackages -scoop_path $scoop_path)
```

you ensure that the output of `Get-ScoopPackages` is always an array and never unrolled. Once you get CI green I'm happy to merge this in.

JamieMagee:
Failing on Windows 2012. Is that VM running an older version of PowerShell? I can't tell from the logs.

jborean93:
Server 2012 will be running with PowerShell v3 which is the minimum version of PowerShell that Ansible supports? Does scoop have a required version?

jborean93:
Now that we've released version `0.2.0` of this collection any new bugfix of feature (excluding completely new plugins) should have a changelog fragment under https://github.com/ansible-collections/community.windows/tree/main/changelogs/fragments.

In this case you could have a changelog fragment with something like (feel free to change the actual entry wording)

```yaml
bugfixes:
- win_scoop - add checks for globally installed packages for better idempotency checks
```

JamieMagee:
Thanks for the pointer. I didn't realise `.Where` was PowerShell v4+.

I think this is ready for merge now.

changelogs/changelog.yaml
```diff
@@ -26,6 +26,7 @@ releases:
       - win_iis_website - Actually restart the site when ``state=restarted`` - https://github.com/ansible/ansible/issues/63828
       - win_partition - Fix invalid variable name causing a failure on checks - https://github.com/ansible/ansible/issues/62401
       - win_partition - don't resize partitions if size difference is < 1 MiB
+      - win_scoop - add checks for globally installed packages for better idempotency checks
```
jborean93, line 29: This needs to be in a separate file in `changelogs/fragments/<name>.yml`. The `changelog.yaml` file is auto generated when the next release is created based on the fragments that are present.

jborean93:
Looks good, the changelog entry just needs to be a new fix so we can denote the fix was in the upcoming release and not an existing one.

JamieMagee:
Thanks. I've moved it now

Issue #78, opened 2020-04-20
Add SRV Record to win_dns_record Module


CSimpiFoN:
##### SUMMARY
It would be great to manage SRV records too on Windows DNS server with win_dns_record module

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
win_dns_record

##### ADDITIONAL INFORMATION
Extend the functionality of win_dns_module

```
- name: SRV example
  win_dns_record:
    name: "metrics.node"
    type: "SRV"
    value: "db1"
    port: "9100"
    zone: "example.com"
```

sakar97:
Working on it.

jborean93:
Was implemented by https://github.com/ansible-collections/community.windows/pull/112 and included in the 1.0.0 release of this collection.

PR #112, opened 2020-07-13
New Feature SRV record in win_dns_record module

sakar97:
##### SUMMARY

#78
Added a new feature of SRV dns record to an existing win_dns_record module.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

type: "SRV"
- It will create an SRV resource record or will remove it.

##### ADDITIONAL INFORMATION

With type: "SRV" three parameters are also created which are:
1- port: - specify the port number to the SRV record.
2- priority: - prioritize the multiple services given by SRV record.
3- weight: - If priority of two records is the same then it will differ by weight. 

- name: Creating a SRV record with port number and priority
    name: "test"
    priority: 5
    port: 995
    weight: 2
    state: present
    type: "SRV"
    values: "amer.example.com"
    zone: "example.com"

```paste below

```


plugins/modules/win_dns_record.ps1
```diff
@@ -123,8 +130,20 @@ if ($null -ne $values -and $values.Count -gt 0) {
...
+                       TimeToLive = $ttl
+        }
         try {
-            Add-DnsServerResourceRecord -ZoneName $zone -Name $name -AllowUpdateAny -TimeToLive $ttl @splat_args -WhatIf:$module.CheckMode @extra_args
+            if ($type -ne 'SRV') {
+                Add-DnsServerResourceRecord -ZoneName $zone -Name $name -AllowUpdateAny -TimeToLive $ttl @splat_args -WhatIf:$module.CheckMode @extra_args
+            }
+            else {
+                Add-DnsServerResourceRecord -SRV -Name $name @srv_args @extra_args -WhatIf:$module.CheckMode
+            }
```
jborean93, line 146: It's a bit easier to maintain code when the positive condition is being checked instead of the inverse, i.e.

```suggestion
            if ($type -eq 'SRV') {
                Add-DnsServerResourceRecord -SRV -Name $name @srv_args @extra_args -WhatIf:$module.CheckMode
            }
            else {
                Add-DnsServerResourceRecord -ZoneName $zone -Name $name -AllowUpdateAny -TimeToLive $ttl @splat_args -WhatIf:$module.CheckMode @extra_args
            }
```

plugins/modules/win_dns_record.py
```diff
@@ -19,6 +19,20 @@
     - The name of the record.
     required: yes
     type: str
+  port:
+    description:
+    - The port number of the record.
+    - Optional for C(type=SRV).
```
jborean93, line 25: It doesn't seem to be optional, the docs for `Add-DnsServerResourceRecord` show it as a mandatory parameter.,

plugins/modules/win_dns_record.py
```diff
@@ -19,6 +19,20 @@
...
+    - Supported only for C(type=SRV).
+    default: ""
```
jborean93, line 27: Default values for int types should be an actual number not an empty string.

plugins/modules/win_dns_record.py
```diff
@@ -19,6 +19,20 @@
...
+    type: int
+  priority:
+    description:
+    - The priority number for each service in SRV record.
+    - Optional for C(type=SRV).
+    - Supported only for C(type=SRV).
+    default: "0"
```
jborean93, line 34: No need to quote an integer

plugins/modules/win_dns_record.ps1
```diff
@@ -8,10 +8,13 @@
 $spec = @{
     options = @{
         name = @{ type = "str"; required = $true }
+        port = @{ type = "int"; default = [int]"" }
```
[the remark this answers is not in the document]
sakar97: Yes, `-Port` `-Priority` and `-Weight` is required parameters that's why I set their default values as 0,0 and an empty string for the port because Port 0 is invalid. If users don't want to mention weigh, priority, or port to their record then also it should work with default values.

plugins/modules/win_dns_record.py
sakar97: Yes, as I have told you above we are not enforcing the user to provide these values instead we are taking it as default.

plugins/modules/win_dns_record.py
sakar97: I thought for `-Port` value 0 is not making any sense because Port 0 is invalid.

plugins/modules/win_dns_record.ps1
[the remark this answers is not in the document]
jborean93: You want to enforce the user to specify these when `type: SRV` so while they aren't required in all cases they should be required in that case. Setting the default values like it is in the PR currently is not a good idea, especially because a port of 0 doesn't make any sense. The suggestion is to remove the defaults and use that `required_if` value to have it mandatory when `type: SRV` but not in other cases.

plugins/modules/win_dns_record.py
jborean93: That's why it doesn't make sense having that as the default value, it shouldn't have any defaults and just be "required" if `type: SRV`. That can't be documented with `required = $true` but rather in the `description` for the option and using that `required_if` statement I had written.

sakar97: Okay Understood

plugins/modules/win_dns_record.py
sakar97: when I was writing `default: 0` it was causing the mismatch with the ps1 file. The actual error was "doc-default-does-not-match-spec ". In the doc default, it is taking it as None and in the ps1 file, I assigned it as 0.

jborean93: The default value type should match the type of the option which in this case is an int. So the default in the doc should be `default: 0` and the default value in the arg spec is `default = 0`.

sakar97: Yes, I tried exactly same in the first commit and facing the "doc-default-does-not-match-spec" error. Which is why I quote the above integer.

sakar97:
@jborean93 can I get a review?

plugins/modules/win_dns_record.py
sakar97: Fixed the changes

jborean93:
Sorry for the delay, the changes look good.

PR #117, opened 2020-07-22
Initial ipv6 enable/disable module

jirolin:
##### SUMMARY

Initial implementation of ```win_ipv6``` module

This module is a function to disable or enable ipv6 status of Windows network adapter.

Most Japanese companies often disable ipv6 on Windows OS network adapters.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

win_ipv6


jborean93:
So while I do see a use case for this I don't think we want just a module to deal with the actual state of 1 network adapter component. It sounds like we want something like `win_net_adapter_feature` that can be used to configure the features of a network adapter like IPv6 and so on. It could be limited to just IPv6 at the start but it allows it to be expanded in the future with more features as desired.

jirolin:
@jborean93 
Thanks for your comment!
It seems good that we add args.componentId.
When I want to enable or disable the IPv6 network adapter,
I can specify the componentId as ms_tcpip6.
If I want to enable or disable another network adapter,
I can specify the componentId as ms_tcpip6.ms_lldp,ms_rspndr, ms_implat, and so on.
The "win_net_adapter_feature" is wonderful name!
So, we want to adopt this name for this new win_module name.

I will try using the following.

```
---
- win_net_adapter_feature:
    interfaces:
      - Ethernet0
      - Ethernet1
    status: disable
    componentID:
      - ms_tcpip6
      - ms_implat
```

jborean93:
I would even suggest further that we make `componentID` to something like `component` or even just `feature` to fix the module name. There should be a module maintained mapping between a human friendly name like `ipv6` to `ms_tcpip6` so end users of the module don't need to know the actual component name as set in Windows. Seeing something like

```
- win_net_adapter_feature:
    interfaces:
      - Ethernet0
      - Ethernet1
    state: disable
    feature:
      - ipv6
      - multiplexor
```

Makes it a bit more obvious we are disabling the `ipv6` and `multiplexor` features/components for those adapters. Also using `state` is more aligned with what most of our other modules support.

jirolin:
@jborean93 
Thanks for your suggestion!

Just as you say, it's good that mapping `componentID` to `feature` and adoption of  `state` when we consider the user-friendliness of this module.
I'm just concerned the following two cases.

1. We may not be able to all of the `componentID`. 
Firstly, when we mapping `componentIDs` , I think it's hard to support all of the `componentIDs`.
For example, in the following case(my Windows10),
there are two special network components.("Trend Micro NDIS 6.0 Filter Driver" and "Juniper Network Service") 
I wonder if the best way is to map only Windows default `componentIDs`,
 or not to map, and just to specify `componentID` whatever each user selects like "Tmlwf"and "jnprns".
```
PS C:\Users\jirolin> Get-NetAdapterBinding |where-object {$_.Name -match "Wi-Fi"}| select DisplayName, ComponentID      
DisplayName                                       ComponentID
-----------                                       -----------
Microsoft LLDP プロトコル ドライバー              ms_lldp
Link-Layer Topology Discovery Responder           ms_rspndr
Microsoft Network Adapter Multiplexor Protocol    ms_implat
インターネット プロトコル バージョン 4 (TCP/IPv4) ms_tcpip
Trend Micro NDIS 6.0 Filter Driver                Tmlwf
Microsoft ネットワーク用クライアント              ms_msclient
Microsoft ネットワーク用ファイルとプリンター共有  ms_server
Link-Layer Topology Discovery Mapper I/O Driver   ms_lltdio
インターネット プロトコル バージョン 6 (TCP/IPv6) ms_tcpip6
QoS パケット スケジューラ                         ms_pacer
Juniper Network Service                           jnprns
```

2. the case of the win_feature module
Secondly, in the [win_feature module](https://docs.ansible.com/ansible/latest/modules/win_feature_module.html),
there are no mapped `features`, we just specifiy names of `features`.
I just know that the number of `features`  supported by the `win_feature` module is much larger than that of `componentID`,
so it's difficult to map all of the `features`.
And,I know we may not need to map the `features` ,because it's easier to understand the name of `features` than the name of `coponentID`.

I'd like to hear your opinion based on the cases above.

jborean93:
I suppose if there was a simple way of translating a name to the actual ID in PowerShell that's less of an issue. From my perspective when I looked at your original example I had to look up what `ms_implat` represented. This somewhat goes against the self documenting nature of an Ansible task but I can see how it doesn't mesh well with custom adapters. I think for now lets just go with the raw component name like you originally had and make sure we put in the docs for the module a way to translate the name in the GUI to the component name.

The option name should also follow the `snake_case` format but in reality I would just have it as `component` or `feature`.

jirolin:
@jborean93 
Thanks for your comment!

I modified the `win_net_adapter_feature.py` in order to make it easier to confirm `componentID`s.
So, we specify `componentID`s ,but we can confirm the meaning that the `componentID`s imply.
I would like you to check this module.

plugins/modules/win_net_adapter_feature.ps1
```diff
@@ -0,0 +1,127 @@
...
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible. If not, see <http://www.gnu.org/licenses/>.
+#
```
jborean93, line 19:
```suggestion

# Copyright: (c) 2020, ライトウェルの人 <jiro.higuchi@shi-g.com>
# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)

```

We try to use the shorter copyright text.

plugins/modules/win_net_adapter_feature.py
```diff
@@ -0,0 +1,70 @@
...
+      - ms_rspndr (Link-Layer Topology Discovery Responder)
+      - ms_msclient (Client for Microsoft Networks)
+      - ms_pacer (QoS Packet Scheduler)
+      - If you'd like to set custom adapters like 'Juniper Network Service',
+      - confirm the componentID by 'Get-NetAdapterBinding' cmdlet.
+    type: str
+    required: yes
+
+author:
+  - ライトウェルの人 (@jirolin)
```
jborean93, line 48: There's no need to change this, it will be interesting to see if our build tools will work with non-ascii characters. It should be fine but merely just a comment that I haven't seen a module use Japanese characters before.

plugins/modules/win_net_adapter_feature.py
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2020, ライトウェルの人 <jiro.higuchi@shi-g.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+# this is a windows documentation stub.  actual code lives in the .ps1
+# file of the same name
```
jborean93, line 8: No need for these lines anymore

```suggestion
```

plugins/modules/win_net_adapter_feature.py
```diff
@@ -0,0 +1,70 @@
...
+
+DOCUMENTATION = r'''
+---
+module: win_net_adapter_feature
+short_description: Enable or disable certain network adapters.
```
jborean93, line 13:
```suggestion
short_description: Enable or disable certain network adapters features.
```

plugins/modules/win_net_adapter_feature.py
```diff
@@ -0,0 +1,70 @@
...
+description:
+     - Enable or disable ms_tcpip6 of a certain network adapter or all the network adapters.
```
jborean93, line 15: The module is no longer limited to just IPv6, would be best to make this a bit more generic.,

plugins/modules/win_net_adapter_feature.py
```diff
@@ -0,0 +1,70 @@
...
+options:
+  interface:
+    description:
+      - Name of Network Adapter Interface. For example, 'Ethernet0' or '*'.
```
jborean93, line 19: When quoting values, best to use `C()` instead of `''`. Using `C()` will have our doc generator emphasis the text and it's designed to be used for module option values.

```suggestion
      - Name of Network Adapter Interface. For example, C(Ethernet0) or C(*).
```

Another common formatter to use is `I()` which is designed to be used with module options themselves, like `I(state)` and so on.

plugins/modules/win_net_adapter_feature.py
```diff
@@ -0,0 +1,70 @@
...
+    type: str
+    choices:
+      - enable
+      - disable
+    required: yes
+  componentID:
+    description:
+      - Specify the below componentID of network adapters.
+      - componentID (DisplayName)
+      - ms_implat (Microsoft Network Adapter Multiplexor Protocol)
```
jborean93, line 34: I would enclose the actual `componentID` in `C()` like so

```suggestion
      - C(ms_implat) (Microsoft Network Adapter Multiplexor Protocol)
```

plugins/modules/win_net_adapter_feature.py
```diff
@@ -0,0 +1,70 @@
...
+      - ms_lltdio (Link-Layer Topology Discovery Mapper I/O Driver)
+      - ms_tcpip6 (Internet Protocol Version 6 (TCP/IPv6))
+      - ms_tcpip (Internet Protocol Version 4 (TCP/IPv4))
+      - ms_lldp (Microsoft LLDP Protocol Driver)
+      - ms_rspndr (Link-Layer Topology Discovery Responder)
+      - ms_msclient (Client for Microsoft Networks)
+      - ms_pacer (QoS Packet Scheduler)
+      - If you'd like to set custom adapters like 'Juniper Network Service',
+      - confirm the componentID by 'Get-NetAdapterBinding' cmdlet.
```
jborean93, line 43: When you add a new line with `-` you are putting the values in separate paragraphs. Just continue the line as the limit is ~160 characters

```suggestion
      - If you'd like to set custom adapters like 'Juniper Network Service', get the I(component_id) by running the C(Get-NetAdapterBinding) cmdlet.
```

plugins/modules/win_net_adapter_feature.py
```diff
@@ -0,0 +1,70 @@
...
+    required: yes
+  state:
+    description:
+      - Specify the state of ms_tcpip6 of interfaces.
+    type: str
+    choices:
+      - enable
+      - disable
+    required: yes
+  componentID:
```
jborean93, line 30: For Ansible options we follow the `snake_case` format. This means this should be `component_id`.

```suggestion
  component_id:
```

plugins/modules/win_net_adapter_feature.py
```diff
@@ -0,0 +1,70 @@
...
+    required: yes
+
+author:
+  - ライトウェルの人 (@jirolin)
+'''
+
+
+EXAMPLES = r'''
+- name: enable multiple interfaces of multiple interfaces
+  win_net_adapter_feature:
```
jborean93, line 54: Since moving to a collection the examples should be the fully qualified collection name like so

```suggestion
  community.windows.win_net_adapter_feature:
```

plugins/modules/win_net_adapter_feature.py
```diff
@@ -0,0 +1,70 @@
...
+    interface:
+    - 'Ethernet0'
+    - 'Ethernet1'
+    state: enable
+    componentID:
+    - ms_tcpip6
+    - ms_server
+
+- name: Enable ms_tcpip6 of all the Interface
+  win_net_adapter_feature:
```
jborean93, line 64:
```suggestion
  community.windows.win_net_adapter_feature:
```

plugins/modules/win_net_adapter_feature.ps1
```diff
@@ -0,0 +1,127 @@
...
+    If($componentID_name.Length -gt 0) {
+        $componentIDs = @($componentID_name)
+    }
+    Else {
+        $componentIDs = @()
+    }
+}
+Else {
+    $componentIDs = $componentID_name
+}
```
jborean93, line 75: So we want to avoid adding any new modules that use the `Legacy` wrapper. There are quite a few missing features when you use this util compared to the replacement. Also when you use the replacement you can get rid of a lot of boilerplate checks that are currently present in the module like so.

Keep in mind there are a few things you need to do differently compared to `Ansible.ModuleUtils.Legacy` like:

* Any calls to `Fail-Json` need to be replaced with `$module.FailJson("msg")`
* Any calls to `Exit-Json` need to be replaced with `$module.ExitJson()`
* Setting a `$result.changed = $true` is now `$module.Result.changed = $true`

```suggestion
$spec = @{
    options = @{
        interface = @{ type = 'list'; elements = 'str'; required = $true }
        state = @{ type = 'str'; choices = 'disabled', 'enabled'; default = 'enabled' }
        component_id = @{ type = 'list'; elements = 'str'; required = $true }
    }
    supports_check_mode = $true
}
$module = [Ansible.Basic.AnsibleModule]::Create($args, $spec)

$interface = $module.Params.interface
$state = $module.Params.state
$component_id = $module.Params.component_id
$check_mode = $module.CheckMode
```

plugins/modules/win_net_adapter_feature.ps1
```diff
@@ -0,0 +1,127 @@
...
+
+Try {
```
jborean93, line 77: Don't wrap the code in a blank `try/catch`. Any unhandled exceptions are reported back to Ansible and by catching and calling Fail-Json you are effectively loosing some information that might be useful to the end user.

jborean93 (changes requested):
I've left some comments in the review. There are 2 major things that need to be done before this module can be merged

* Move away from `Ansible.ModuleUtils.Legacy` to the newer format
* Add tests for check mode operations

The first brings a lot of benefits that are missing with the `Legacy` wrapper like

* Doc validation against the actual argspec
* Argspec value reporting when running with `-vvv` at runtime
* Better value validation, support for list values
* Better error reporting with `FailJson`

Happy to help with the transition where I can but hopefully the comments I've left so far help start the process. You can also have a read of https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_general_windows.html to learn more about this newer format.

plugins/modules/win_net_adapter_feature.py
jirolin: Thank you for pointing it out!
I missed it.

plugins/modules/win_net_adapter_feature.py
jirolin: Thanks for the description fix!

plugins/modules/win_net_adapter_feature.py
jirolin: Thank you for pointing it out!

plugins/modules/win_net_adapter_feature.py
jirolin: Thank you!

plugins/modules/win_net_adapter_feature.py
jirolin: Thank you!

plugins/modules/win_net_adapter_feature.py
jirolin: As you say, it is the rare case that  a module use Japanese characters.
I would like to leave my account name as it is.

plugins/modules/win_net_adapter_feature.py
jirolin: Thank you.
I need to learn a lot of things for transition to Ansible.collection.
I hadn't checked it thoroughly.

plugins/modules/win_net_adapter_feature.ps1
jirolin: Thank you for pointing it out!
I need to fix this and to execute tests again for transition to ansible collection.

plugins/modules/win_net_adapter_feature.ps1
jirolin: Thank you!
I need to check the docs carefully and to execute tests again.

plugins/modules/win_net_adapter_feature.py
jborean93: That's perfectly fine, sorry I didn't mean to say you should change it. I was merely trying to say that I've never encountered it before and I hope the doc builder can handle it fine (which it should) :).

plugins/modules/win_net_adapter_feature.ps1
jborean93: No worries, it unfortunately isn't a simple swap over but we tried to make it as painless as we could while adding support for newer features.

Hax7:
I would like to see this module merged and I hope interfaces can be set to all like in this command `"Disable-NetAdapterBinding -Name "*" -ComponentID ms_tcpip6"`

jirolin:
@Hax7 
Thanks for your comment!
As the following example, the module can accept `'*'` as the `interface` argument.

```
- name: Enable ms_tcpip6 of all the Interface
  community.windows.win_net_adapter_feature:
    interface: '*'
    state: enabled
    component_id:
    - ms_tcpip6
```
@jborean93 
After merging your improved code, I've improved the code and the test cases as possible.
I would be grateful if you could check this module again.

jborean93 (approved):
Sorry for the delay, the changes look great. I've just made 2 simple changes so we can merge it in

1. Added `version_added: 1.2.0` to the module docs as that will be the first version it will appear in this collection
2. Removed the overall try/catch block in the module code. This makes it quite hard to pick up the exception details and the PowerShell wrapper does a better job in logging these things if they are uncaught.

For the 2nd, it's best to only use try/catch if there's a specific exception you want to handle, or you know the error messages are a bit vague and can be improved.

jborean93:
It turned out the tests weren't running in CI and required a certain setup on the host to run. I've also updated them to be generic and set them to run in CI.

jborean93:
Thanks for working on this one, the changes you've made have been great and it will nice to have this in the upcoming release of the collection.

jirolin:
@jborean93 
I'm grateful for your help and merging.
I wish I can contribute some modules to this project again.
See you again.

Issue #13, opened 2020-03-11
win_firewall_rule Allow enabling built-in rules by using displaygroup


jborean93:
_From @kinwolfqc on Mar 05, 2020 19:28_

##### SUMMARY

In powershell, we can enable multiple rules that belong to the same displaygroup, and  many built-in rules use disaplygroups  
`Enable-NetFirewallRule -DisplayGroup "Windows Remote Management"`

The win_firewall_rule doesn't seems to allow that.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

win_firewall_rule

_Copied from original issue: ansible/ansible#68052_

sakar97:
@jborean93  can I work on this?

jborean93:
Sure, any PR will be appreciated.

sakar97:
I came up with two Scenarios:
1- When a user is passing only the `group` and `enabled` parameter. Then it will enable all the rules in the provided group.
2- When a user is passing only the `group` and a new parameter which is something like `enableall`. Then it will enable all the rules in the group.
In both cases, I need to change the `name` parameter as optional, otherwise it will create, modify, or remove the named rule.
Any suggestions will be helpful and appreciated in implementing this new feature.

jborean93:
I would probably prefer option 1, by passing a `group` and not `name` you are effectively saying this is the state for all rules in this group.

PR #127, opened 2020-08-11
New Feature enable all FW rules in a group in win_firewall_rule module

sakar97:
##### SUMMARY

Added a new feature that allows the user to enable or disable all the firewall rules in a particular group.

Fixes https://github.com/ansible-collections/community.windows/issues/13

##### ISSUE TYPE

- Feature Pull Request
##### ADDITIONAL INFORMATION

When a user inputs only the group and enabled parameter than it will enable all the rules in a group.  
Removed `name` from the required parameter.

```paste below
- name: Enable all the Firewall rules in the application group
  win_firewall_rule:
    group: application
    enabled: yes
```

### DIFF
--- a/changelogs/fragments/win_firewall_rule_group_state.yml
+++ b/changelogs/fragments/win_firewall_rule_group_state.yml
@@ -1,0 +1,3 @@
+minor_changes:
+- win_firewall_rule - Support editing rules that have a duplicated name
+- win_firewall_rule - Support editing rules by the group it belongs to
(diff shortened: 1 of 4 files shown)


jborean93:
Thanks @sakar97 for your work here, I've pushed a commit that takes the next step forward and allow setting other properties based on the group name and not just `enabled`. I've also slightly tweaked the doc changes and added a changelog fragment.

PR #130, opened 2020-08-15
New Module: win_domain_ou

zollo:
##### SUMMARY
The win_domain_ou module manages Active Directory Organizational Units (OUs).

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
win_domain_ou

##### ADDITIONAL INFORMATION


plugins/modules/win_domain_ou.ps1
```diff
@@ -0,0 +1,197 @@
...
+        if ($Object.properties.city) { $parms.City = $Object.properties.city }
+        if ($Object.properties.state) { $parms.State = $Object.properties.state }
+        if ($Object.properties.street_address) { $parms.StreetAddress = $Object.properties.street_address }
+        if ($Object.properties.postal_code) { $parms.PostalCode = $Object.properties.postal_code }
+        if ($Object.properties.country) { $parms.Country = $Object.properties.country }
+        if ($Object.properties.managed_by) { $parms.ManagedBy = $Object.properties.managed_by }
+    }
+
+    # convert to psobject & return
+    return New-Object psobject -Property $parms
```
jborean93, line 79: Just a nitpick here but you can just use the `[PSCustomObject]` "cast"

```suggestion
    [PSCustomObject]$parms
```

plugins/modules/win_domain_ou.ps1
```diff
@@ -0,0 +1,197 @@
...
+        Try { New-ADOrganizationalUnit @parms -ProtectedFromAccidentalDeletion $protected -WhatIf:$check_mode }
+        Catch { $module.FailJson("Failed to create organizational unit: $($_.Exception.Message)") }
+    }
+
+    # ou exists, update object
+    if ($current_ou) {
+        Try { Set-ADOrganizationalUnit -Identity "OU=$name,$path" @parms -WhatIf:$check_mode }
+        Catch {
+            $module.Result.debug = $parms
+            $module.FailJson("Failed to update organizational unit: $($_.Exception.Message)") }
```
jborean93, line 142: Should add `$_` as a 2nd argument to `FailJson` so the stacktrace isn't lost when using a higher verbosity in Ansible.

plugins/modules/win_domain_ou.py
```diff
@@ -0,0 +1,174 @@
...
+        l(protected=true), you cannot delete the corresponding object without
+        changing the value of the property.
+    type: bool
+    default: false
+  path:
+    description:
+      - Specifies the X.500 path of the OU or container where the new object is
+        created.
+    type: str
+    required: true
```
jborean93, line 37: I feel like the path shouldn't be required, couldn't we have the default as the base OU for the domain? I'm not too familiar with best practices here so please correct me if I'm wrong.

plugins/modules/win_domain_ou.py
```diff
@@ -0,0 +1,174 @@
...
+    type: str
+    default: present
+    choices: [ present, absent ]
+  recursive:
+    description:
+      - Removes the OU and any child items it contains.
+      - You must specify this parameter to remove an OU that is not empty.
+    type: bool
+    default: false
+  properties:
```
jborean93, line 55: Should be a custom list of properties we know off or just directly mapped to the LDAP property name. We do the latter for some other domain modules in Ansible and it makes it more flexible in the future.

plugins/modules/win_domain_ou.py
```diff
@@ -0,0 +1,174 @@
...
+      - Specifies the desired state of the OU.
+      - When l(state=present) the module will attempt to create the specified
+        OU if it does not already exist.
+      - When l(state=absent), the module will remove the specified OU.
+      - When l(state=absent) and l(recursive=true), the module will remove all
+        the OU and all child OU's.
+    type: str
+    default: present
+    choices: [ present, absent ]
+  recursive:
```
jborean93, line 49: I'm torn with this one, we don't need to specify `recursive: yes` for `win_file` when deleting a directory. Was there a particular reason why you you decided to have this as an opt-in behaviour?

jborean93 (commented):
Just had a brief skim and left some comments inline. It would also be nice to support a `domain_server`, `domain_username`, and `domain_password` options to provide a way to set them explicitly.

plugins/modules/win_domain_ou.py
joezollo: Agreed! Will get this implemented!

plugins/modules/win_domain_ou.py
joezollo: I think this is a great idea, if the user doesn't provide a path we can just use the base path for the domain and create objects there. So if we want to create a new OU called "AnsibleFest" in the domain test.ansible.com, our distinguished name would be **OU=AnsibleFest,DN=test,DN=ansible,DN=com**

```yaml
- name: Ensure OU is present
  community.windows.win_domain_ou:
    name: AnsibleFest
```

plugins/modules/win_domain_ou.py
joezollo: This was done mainly out of caution - I wanted to be absolutely sure that the intent of the user is to perform a recursive delete operation. I'm not opposed to changing it at all.

jborean93: I'll bring it up in the next windows working group and see what the general consensus is on this.

jborean93: So the windows working group decided on not having this option and to always use the `-Recursive` switch on the remove parameter.

joezollo: @jborean93 Sounds good, i'll kill this option - refactor in progress!

git-cgallagher: Would we want to kill this option or just default it to true?

dcardellino:
any update on this? would be very useful for our company to have that feature.

zollo:
> any update on this? would be very useful for our company to have that feature.

I'll have some time to finish this up in the coming week or so. Sorry! The module works fine now but I want to have it align better with Ansible's design.

plugins/modules/win_domain_ou.py
gamethis: I would say you just set the default to True if that is what the group wanted... though the safe way would be to default to False.  
In large environments with nested OU's deleting everything recursively by default sounds like a bad idea.   but having the option to not do recursive allows the user to make the choice.

gamethis:
if @zollo is not going to finish this up ... I would be happy to give it a second go.
I need some updates for it to work... and can incorporate those in.

jborean93:
Closing due to the age of the PR. If someone wishes to continue with the work they can submit a new PR with the changes requested added in.

Issue #133, opened 2020-08-20
Inbound and outbound action in win_firewall module.


sakar97:
##### SUMMARY

Currently, the win_firewall module does not manage inbound and outbound action, whether allow inbound rules or block them and similar for outbound actions.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

win_firewall
##### ADDITIONAL INFORMATION

```yaml
- name: Enable firewall for Domain, Public and Private profiles and block outbound actions
  community.windows.win_firewall:
    state: enabled
    profiles:
    - Domain
    - Private
    - Public
    inbound: allow
    outbound: block
  tags: enable_firewall
```

jborean93:
Was implemented with https://github.com/ansible-collections/community.windows/pull/136.

PR #136, opened 2020-08-27
New feature support inbound,outbound connection in win_firewall module.

sakar97:
##### SUMMARY

#133 
Added a new feature of inbound, outbound connections that allow the user to either block or allow the connection in windows firewall.

##### ISSUE TYPE

- Feature Pull Request
##### ADDITIONAL INFORMATION

Added two parameters with inbound, outbound which takes the value as allow and block.

```paste below
- name: Enable firewall for Domain profile and block outbound connections
  win_firewall:
    profiles: Domain
    state: enabled
    outbound: block
  tags: block_connection
```

### DIFF
--- a/changelogs/fragments/win_firewall_action.yml
+++ b/changelogs/fragments/win_firewall_action.yml
@@ -1,0 +1,2 @@
+minor_changes:
+- win_firewall - Support inbound, outbound connection in windows firewall.
--- a/changelogs/fragments/win_firewall_action.yml
+++ b/changelogs/fragments/win_firewall_action.yml
@@ -1,3 +1,2 @@
 minor_changes:
-- win_firewall - Support inbound, outbound connection in windows firewall.
-- win_firewall - By default inbound connections are block and outbound connections are allow.
+- win_firewall - Support defining the default inbound and outbound action of traffic in Windows firewall
--- a/changelogs/fragments/win_firewall_action.yml
+++ b/changelogs/fragments/win_firewall_action.yml
@@ -1,2 +1,3 @@
 minor_changes:
 - win_firewall - Support inbound, outbound connection in windows firewall.
+- win_firewall - By default inbound connections are block and outbound connections are allow.
--- a/changelogs/fragments/win_firewall_action.yml
+++ b/changelogs/fragments/win_firewall_action.yml
@@ -1,2 +1,2 @@
 minor_changes:
-- win_firewall - Support defining the default inbound and outbound action of traffic in Windows firewall
+- win_firewall - Support defining the default inbound and outbound action of traffic in Windows firewall.
--- a/plugins/modules/win_firewall.ps1
+++ b/plugins/modules/win_firewall.ps1
@@ -13,15 +13,15 @@ $check_mode = Get-AnsibleParam -obj $params -name "_ansible_check_mode" -type "b
 
 $profiles = Get-AnsibleParam -obj $params -name "profiles" -type "list" -default @("Domain", "Private", "Public")
 $state = Get-AnsibleParam -obj $params -name "state" -type "str" -failifempty $true -validateset 'disabled','enabled'
-$inbound = Get-AnsibleParam -obj $params -name "inbound" -type "str" -default "Block"
-$outbound = Get-AnsibleParam -obj $params -name "outbound" -type "str" -default "Allow"
+$inbound_action = Get-AnsibleParam -obj $params -name "inbound_action" -type "str" -default "block" -validateset 'allow','block'
+$outbound_action = Get-AnsibleParam -obj $params -name "outbound_action" -type "str" -default "allow" -validateset 'allow','block'
 
 $result = @{
     changed = $false
     profiles = $profiles
     state = $state
-    inbound = $inbound
-    outbound = $outbound
+    inbound_action = $inbound_action
+    outbound_action = $outbound_action
 }
 
 try {
@@ -36,9 +36,10 @@ Try {
 
     ForEach ($profile in $firewall_profiles) {
 
-        $currentstate = (Get-NetFirewallProfile -Name $profile).Enabled
-        $current_inboundaction = (Get-NetFirewallProfile -Name $profile).DefaultInboundAction
-        $current_outboundaction = (Get-NetFirewallProfile -Name $profile).DefaultOutboundAction
+        $current_profile = Get-NetFirewallProfile -Name $profile
+        $currentstate = $current_profile.Enabled
+        $current_inboundaction = $current_profile.DefaultInboundAction
+        $current_outboundaction = $current_profile.DefaultOutboundAction
         $result.$profile = @{
             enabled = ($currentstate -eq 1)
             considered = ($profiles -contains $profile)
@@ -56,14 +57,14 @@ Try {
                 $result.changed = $true
                 $result.$profile.enabled = $true
             }
-            if ($inbound -ne $current_inboundaction) {
-                Set-NetFirewallProfile -name $profile -DefaultInboundAction $inbound
-                $result.inbound = $inbound
+            if ($inbound_action -ne $current_inboundaction) {
+                Set-NetFirewallProfile -name $profile -DefaultInboundAction $inbound_action -WhatIf:$check_mode
+                $result.inbound_action = $inbound_action
                 $result.changed = $true
             }
-            if ($outbound -ne $current_outboundaction) {
-                Set-NetFirewallProfile -name $profile -DefaultOutboundAction $outbound
-                $result.outbound = $outbound
+            if ($outbound_action -ne $current_outboundaction) {
+                Set-NetFirewallProfile -name $profile -DefaultOutboundAction $outbound_action -WhatIf:$check_mode
+                $result.outbound_action = $outbound_action
                 $result.changed = $true
             }
 
--- a/plugins/modules/win_firewall.ps1
+++ b/plugins/modules/win_firewall.ps1
@@ -13,8 +13,8 @@ $check_mode = Get-AnsibleParam -obj $params -name "_ansible_check_mode" -type "b
 
 $profiles = Get-AnsibleParam -obj $params -name "profiles" -type "list" -default @("Domain", "Private", "Public")
 $state = Get-AnsibleParam -obj $params -name "state" -type "str" -failifempty $true -validateset 'disabled','enabled'
-$inbound_action = Get-AnsibleParam -obj $params -name "inbound_action" -type "str" -validateset 'allow','block', 'not_configured'
-$outbound_action = Get-AnsibleParam -obj $params -name "outbound_action" -type "str" -validateset 'allow','block', 'not_configured'
+$inbound_action = Get-AnsibleParam -obj $params -name "inbound_action" -type "str" -validateset 'allow','block','not_configured'
+$outbound_action = Get-AnsibleParam -obj $params -name "outbound_action" -type "str" -validateset 'allow','block','not_configured'
 
 $result = @{
     changed = $false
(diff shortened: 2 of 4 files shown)


plugins/modules/win_firewall.ps1
```diff
@@ -33,6 +37,8 @@ Try {
     ForEach ($profile in $firewall_profiles) {
 
         $currentstate = (Get-NetFirewallProfile -Name $profile).Enabled
+        $current_inboundaction = (Get-NetFirewallProfile -Name $profile).DefaultInboundAction
+        $current_outboundaction = (Get-NetFirewallProfile -Name $profile).DefaultOutboundAction
```
jborean93, line 41: You are calling `Get-NetFirewallProfile` 3 times here, call it once and set each var from there

```suggestion
        $current_profile = Get-NetFirewallProfile -Name $profile
        $currentstate = $current_profile.Enabled
        $current_inboundaction = $current_profile.DefaultInboundAction
        $current_outboundaction = $current_profile.DefaultOutboundAction
```

changelogs/fragments/win_firewall_action.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+- win_firewall - Support inbound, outbound connection in windows firewall.
```
jborean93, line 2: I think this needs to be expanded a bit more to state this is the default action for inbound/outbound connection

plugins/modules/win_firewall.ps1
```diff
@@ -13,11 +13,15 @@ $check_mode = Get-AnsibleParam -obj $params -name "_ansible_check_mode" -type "b
 
 $profiles = Get-AnsibleParam -obj $params -name "profiles" -type "list" -default @("Domain", "Private", "Public")
 $state = Get-AnsibleParam -obj $params -name "state" -type "str" -failifempty $true -validateset 'disabled','enabled'
+$inbound = Get-AnsibleParam -obj $params -name "inbound" -type "str" -default "Block"
```
jborean93, line 16: I would probably lean more towards the option names being `inbound_action` and `outbound_action`. Removes any ambiguity as to what the option is actually for. Also

* The values should be lowercase, i.e. `block` and `allow`
* You should add `-validateset 'allow', 'block` so Ansible validates the input for you.

It also seems like we should have `not_configured` to allow people wanting to use GPO for this object.

plugins/modules/win_firewall.ps1
```diff
@@ -13,11 +13,15 @@ $check_mode = Get-AnsibleParam -obj $params -name "_ansible_check_mode" -type "b
...
+$outbound = Get-AnsibleParam -obj $params -name "outbound" -type "str" -default "Allow"
 
 $result = @{
     changed = $false
     profiles = $profiles
     state = $state
+    inbound = $inbound
```
jborean93, line 23: Same this should match the option name so I would change it to `inbound_action` and `outbound_action`.

plugins/modules/win_firewall.ps1
```diff
@@ -50,6 +56,16 @@ Try {
                 $result.changed = $true
                 $result.$profile.enabled = $true
             }
+            if ($inbound -ne $current_inboundaction) {
+                Set-NetFirewallProfile -name $profile -DefaultInboundAction $inbound
```
jborean93, line 60: Need to add `-WhatIf:$check_mode`

plugins/modules/win_firewall.ps1
```diff
@@ -50,6 +56,16 @@ Try {
...
+                $result.inbound = $inbound
+                $result.changed = $true
+            }
+            if ($outbound -ne $current_outboundaction) {
+                Set-NetFirewallProfile -name $profile -DefaultOutboundAction $outbound
```
jborean93, line 65: Need to add `-WhatIf:$check_mode`

jborean93 (changes requested):

plugins/modules/win_firewall.py
```diff
@@ -46,6 +58,12 @@
     profiles:
     - Domain
   tags: disable_firewall
+- name: Enable firewall for Domain profile and block outbound connections
+  win_firewall:
```
jborean93, line 62: Examples should use the fully qualified module name

```suggestion
  community.windows.win_firewall:
```

plugins/modules/win_firewall.ps1
```diff
@@ -13,11 +13,15 @@ $check_mode = Get-AnsibleParam -obj $params -name "_ansible_check_mode" -type "b
 
 $profiles = Get-AnsibleParam -obj $params -name "profiles" -type "list" -default @("Domain", "Private", "Public")
 $state = Get-AnsibleParam -obj $params -name "state" -type "str" -failifempty $true -validateset 'disabled','enabled'
+$inbound_action = Get-AnsibleParam -obj $params -name "inbound_action" -type "str" -default "block" -validateset 'allow','block'
```
jborean93, line 16: Should we have a default value, right now if people don't define it then the module will block the connection whereas before it just ignored the value. Sounds like we don't want a default and to only do the check if it's defined. We should also add `not_configured` as an option as the underlying cmdlet supports this. You would need a switch to convert the snake_case to the PascalCase the cmdlets expect but that shouldn't be too hard.

plugins/modules/win_firewall.ps1
```diff
@@ -13,11 +13,15 @@ $check_mode = Get-AnsibleParam -obj $params -name "_ansible_check_mode" -type "b
...
+$outbound_action = Get-AnsibleParam -obj $params -name "outbound_action" -type "str" -default "allow" -validateset 'allow','block'
 
 $result = @{
     changed = $false
     profiles = $profiles
     state = $state
+    inbound_action = $inbound_action
```
jborean93, line 23: While you can keep it, I personally try to avoid returning the same data that the user already knows. If the module is converted to the newer Ansible.Basic.AnsibleModule wrapper than module invocation data is added by default. Return values should be reserved for data that is generated by the module or information the user may not know already. Anything else is mostly just historical behaviour before we tried to make that policy.

changelogs/fragments/win_firewall_action.yml
```diff
@@ -0,0 +1,3 @@
...
```
jborean93, line 2: Just keep the 1 entry here, the original comment was around making it a bit more descriptive as to what was actually added. Right now seeing `inbound, outbound connection` is a bit vague in what it does. The entry should state something like

> win_firewall - Support defining the default inbound and outbound action of traffic in Windows firewall

plugins/modules/win_firewall.ps1
sakar97: Yes, but the profile and state are also returning which is passed by the user and the user already knows about it, which is why I returned inbound/outbound actions.  Thanks for the explanation though. Really helpful

plugins/modules/win_firewall.ps1
```diff
@@ -50,7 +55,20 @@ Try {
                 $result.changed = $true
                 $result.$profile.enabled = $true
             }
-
+            if($null -ne $inbound_action) {
+                $inbound_action = (Get-Culture).TextInfo.ToTitleCase(($inbound_action.ToLower() -replace "_", " ")) -replace " ", ""
```
jborean93, line 59: You should do `[Globalization.CultureInfo]::InvariantCulture.TextInfo.ToTitleCase(...)` here and below to ensure it always replaces `not_configured` to the proper value. This ensures that regardless of the actual Locale set the operation will be the same which we know to work. Also you can condense the replace into just 1 call

```suggestion
                $inbound_action = [Globalization.CultureInfo]::InvariantCulture.TextInfo.ToTitleCase($inbound_action.ToLower()) -replace '_', ''
```

plugins/modules/win_firewall.ps1
```diff
@@ -50,7 +55,20 @@ Try {
...
+                if ($inbound_action -ne $current_inboundaction) {
+                  Set-NetFirewallProfile -name $profile -DefaultInboundAction $inbound_action -WhatIf:$check_mode
+                  $result.changed = $true
+                }
+            }
+            if($null -ne $outbound_action) {
+                $outbound_action = (Get-Culture).TextInfo.ToTitleCase(($outbound_action.ToLower() -replace "_", " ")) -replace " ", ""
+                if ($null -ne $outbound_action -and $outbound_action -ne $current_outboundaction) {
```
jborean93, line 67: No need for the `$null -ne $outbound_action` check as you've already got this in the higher if statement

```suggestion
                if ($outbound_action -ne $current_outboundaction) {
```

plugins/modules/win_firewall.py
```diff
@@ -24,6 +24,18 @@
     - Set state of firewall for given profile.
     type: str
     choices: [ disabled, enabled ]
+  inbound_action:
+    description:
+    - Allow or block inbound network traffic in the profile.
+    - NotConfigured is valid when configuring a GPO.
+    type: str
+    choices: [ allow, block, not_configured ]
```
jborean93, line 32: This is a new option and should have `version_added: 1.1.0`

plugins/modules/win_firewall.py
```diff
@@ -24,6 +24,18 @@
...
+  outbound_action:
+    description:
+    - Allow or block inbound network traffic in the profile.
+    - NotConfigured is valid when configuring a GPO.
+    type: str
+    choices: [ allow, block, not_configured ]
```
jborean93, line 38: This is a new option and should have `version_added: 1.1.0`

plugins/modules/win_firewall.py
```diff
@@ -24,6 +24,18 @@
     - Set state of firewall for given profile.
     type: str
     choices: [ disabled, enabled ]
+  inbound_action:
+    description:
+    - Allow or block inbound network traffic in the profile.
```
jborean93, line 29: Make sure the values specified match up with the actual choices. Also whenever you reference a value it should be enclosed in `C()`, e.g. `Set to C(allow) or C(block) inbound network traffic in the profile.`

plugins/modules/win_firewall.py
```diff
@@ -24,6 +24,18 @@
...
+    - NotConfigured is valid when configuring a GPO.
```
jborean93, line 30: Same here should match the actual choice and enclosed with `C()`, e.g. `C(not_configured) is valid when configuring a GPO.`

plugins/modules/win_firewall.py
```diff
@@ -46,6 +58,12 @@
     profiles:
     - Domain
   tags: disable_firewall
+- name: Enable firewall for Domain profile and block outbound connections
```
jborean93, line 61: Make sure you have an empty line between this example and the preview one

sakar97:
Why test cases are getting skipped and for windows 2016/2019 it is unstable?

jborean93:
They are getting skipped for 2012/2012 R2 because that's what the test is configured to do https://github.com/ansible-collections/community.windows/blob/main/tests/integration/targets/win_firewall/aliases#L2-L3. My guess is the cmdlets used in this module are only available on PowerShell 5.1 or Server 2016+.

The tests are unstable because they fail once then on the rerun they work fine. Have a look at the output for one of the tests to see why it's failing https://app.shippable.com/github/ansible-collections/community.windows/runs/536/24/console. We can see this particular task is failing because your assertion isn't correct.

```
03:45 TASK [win_firewall : Turn on Windows Firewall on Domain with block inbound connection] ***
03:48 changed: [windows-2019] => {"Domain": {"considered": true, "currentstate": 1, "enabled": true}, "Private": {"considered": false, "currentstate": 1, "enabled": true}, "Public": {"considered": false, "currentstate": 1, "enabled": true}, "changed": true, "profiles": "Domain", "state": "enabled"}
03:48 
03:48 TASK [win_firewall : Test firewall_domain_on (normal mode)] ********************
03:48 fatal: [windows-2019]: FAILED! => {
03:48     "assertion": "firewall_domain_on is not changed",
03:48     "changed": false,
03:48     "evaluated_to": false,
03:48     "msg": "Assertion failed"
03:48 }
```

sakar97:
Hey, Thanks for the detailed explanation. I miss that error part though and was expecting some other results from the shippable.

jborean93 (approved):
Looks great, thanks for implementing this change.

PR #146, opened 2020-10-03
Fixed requirements for win_scheduled_task

mlschuh:
* Removed unneeded required properties

* Added (already supported) trigger state_change to documentation

##### SUMMARY

Fixes the inability to use `session_state_change` trigger types in scheduled tasks. The module was requiring properties that were not permitted by Powershell. Also added in the documentation for the `state_change` property which, although supported, was not documented.

##### ISSUE TYPE
- Bugfix Pull Request

##### ADDITIONAL INFORMATION

Supporting documentation for the Scheduled Task State Change Trigger: https://docs.microsoft.com/en-us/windows/win32/taskschd/sessionstatechangetrigger

Supporting documentation for the Scheduled Task State Change Trigger expected state change type: https://docs.microsoft.com/en-us/windows/win32/taskschd/sessionstatechangetrigger-statechange

When running the playbook, an error would be thrown that the key was mandatory:
```
fatal: [172.24.0.160]: FAILED! => {"changed": true, "msg": "mandatory key 'days_of_week' for trigger is not set, mandatory keys are 'days_of_week', 'start_boundary'"}
```

But when you included it, Powershell rejected it:
```
fatal: [172.24.0.160]: FAILED! => {"changed": false, "msg": "failed to set trigger property 'DaysOfWeek' to '127': Exception setting \"DaysOfWeek\": \"The property 'DaysOfWeek' cannot be found on this object. Verify that the property exists and can be set.\""}
```

### DIFF
--- a/changelogs/fragments/scheduled_task_state.yml
+++ b/changelogs/fragments/scheduled_task_state.yml
@@ -1,0 +1,3 @@
+minor_changes:
+- win_scheduled_task - Support setting human friendly ``state_change`` values on a ``session_state_change`` trigger
+- win_scheduled_task_state - Added ``state_change_str`` to the trigger output to give a human readable description of the value
--- a/changelogs/fragments/scheduled_task_state.yml
+++ b/changelogs/fragments/scheduled_task_state.yml
@@ -1,3 +1,3 @@
 minor_changes:
-- win_scheduled_task - Support setting human friendly ``state_change`` values on a ``session_state_change`` trigger
+- win_scheduled_task - Added support for setting a ``session_state_change`` trigger by documenting the human friendly values for ``state_change``
 - win_scheduled_task_state - Added ``state_change_str`` to the trigger output to give a human readable description of the value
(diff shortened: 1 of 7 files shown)


jborean93:
Sorry for the delay in the review here.

So while the code is able to support the state_change option it seems like based on your error it's not technically possible unless you modify the code to remove the `days_of_week` mandatory check. If that is the case then I recommend we change `state_change` to be a string type with the choices `console_connect`, `console_disconnect`, `remote_connect`, `remote_disconnect`, `session_lock`, and `session_unlock`. We can have the code inside the module to convert the string values to the int ones used by the COM interface.

This allows us to make it easier for humans to read and write a task for this module without having to rely on the docs to understand that `1` is `TASK_CONSOLE_CONNECT`. Adding a test for this would also be great to ensure our mapping does actually work properly.

Smily0:
Any idea when this may merge? I am running into this problem and was about to submit a bug before I found this pull request.

mlschuh:
It got pushed to my backburner shortly after I submitted the PR especially as this environment is very outside my normal wheelhouse. I'll see if I can make the adjustments this weekend.

jborean93:
I've pushed some changes based on my last review and also added some tests/changelog fragment for the future. Thanks for working on this!

PR #150, opened 2020-10-15
New module win_ntp

### LINKED ISSUE #114: win_ntp module required to configure NTP
##### SUMMARY
Currently there is no ntp module to configure NTP client. Its good to have a **win_ntp** module that can take multiple NTP server IP/FQDN and configure in the windows system.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
win_ntp

##### ADDITIONAL INFORMATION
Currently only win_timezone module is available related to timezone change. But its having no option to configure the NTP  as well. Hence either enhance this module with ntp configuration or make a new module `win_ntp` for ntp configuration.

sakar97:
##### SUMMARY

Manage Windows NTP Server.
Users can perform the following tasks:
- can change the time zone
- pass the multiple peers to the NTP server.
- for NT5DS, can set cross_site_sync_flag
- enable/disable the NTP client in time service
- can select which peers to accept synchronization from
- can synchronize the clock
- can change the settings to default
 

Fix #114 
##### ISSUE TYPE

- New Module Pull Request
##### COMPONENT NAME

ansible.windows.win_ntp
##### ADDITIONAL INFORMATION

- name: Enable npt service with multiple peers to get sync from
  community.windows.win_ntp:
    enabled: true
    peerlist:
      - time.google.com
      - time.yahoo.com
      - time.apple.com
    type: NTP
    sync_clock: true

```paste below

```


mwtrigg:
Nice work and thank you for your contribution!  I feel inspired to build and submit my own module.

I think this module should be renamed win_w32tm.  Perhaps reserve the win_ntp module name for future use; maybe one that configures all NTP client or server settings.

I think that you should also ensure that all references to w32tm/tzutil should include the ".exe".

sakar97:
Hi, @jborean93 can you review this PR.

plugins/modules/win_ntp.ps1
```diff
@@ -0,0 +1,183 @@
+#!powershell
+
+# Copyright: (c) 2020, Sakar Mehra <sakarmehra100@gmail.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+#Requires -Module Ansible.ModuleUtils.Legacy
```
jborean93, line 6: We are trying to avoid using `Ansible.ModuleUtils.Legacy` for any new modules. Using the `Ansible.Basic` C# util offers numerous advantages over the old style like:

* Module invocation reporting
* More validation options, especially for list or dict types
* Enables sanity tests to check that the docs match up with the module options

Let me know if you need any help trying to convert this over to the newer style. If you need to see other modules that use this wrapper I suggest you look at [win_certificate](https://github.com/ansible-collections/community.windows/blob/main/plugins/modules/win_credential.ps1#L9-L53).

plugins/modules/win_ntp.ps1
```diff
@@ -0,0 +1,183 @@
...
+
+$ErrorActionPreference = "Stop"
```
jborean93, line 8: Not needed anymore, this is always set by the PowerShell wrapper that executes modules.

plugins/modules/win_ntp.ps1
```diff
@@ -0,0 +1,183 @@
...
+    SourceNameRaw = $null
+    LastTimeSyncElapsedSeconds = $null
+    LastTimeSynchronizationDateTime = $null
+    StatusRequiredSourceType= $null
+    StatusDateTime= $null
+    StatusLastTimeSynchronization = $null
+}
+
+$current_time_zone= [System.TimeZoneInfo]::Local.Id
+if ($null -ne $time_zone -and $current_time_zone -ne $time_zone) {
```
jborean93, line 37: We already have a timezone module, we shouldn't overload that functionality here.

plugins/modules/win_ntp.py
```diff
@@ -0,0 +1,109 @@
...
+requirements:
+  - This module requires Windows Server 2012R2 or Newer
+description:
+  - Manage Windows NTP Client
+  - Adds, Removes and Modifies NTP Server in windows machine.
+options:
+  peerlist:
+    description:
+      - Fully qualified domain name or IP address from which computer obtain timestamp.
+    type: list
```
jborean93, line 21: Should also add `elements: str` to state it's a list of strings that are accepted.

plugins/modules/win_ntp.py
```diff
@@ -0,0 +1,109 @@
...
+    description:
+      - Ntp configuration are set to default with this option.
+    type: bool
+    default: false
+  cross_site_sync_flags:
+    description:
+      - Determines whether the service chooses synchronization partners outside the domain of the computer.
+      - Thsi value is ignored if l(type=NT5DS) is not set.
+    type: int
+    choices: [ 0, 1, 2 ]
```
jborean93, line 49: Instead of using integer values you should use the string representations and convert them to the proper int values in the module. It's easier for people to understand what they are for if it's a string.

plugins/modules/win_ntp.py
```diff
@@ -0,0 +1,109 @@
...
+    default: UTC
+  factory_default:
+    description:
+      - Ntp configuration are set to default with this option.
+    type: bool
+    default: false
+  cross_site_sync_flags:
+    description:
+      - Determines whether the service chooses synchronization partners outside the domain of the computer.
+      - Thsi value is ignored if l(type=NT5DS) is not set.
```
jborean93, line 47: `This` and use `C(type=NT5DS)`.

plugins/modules/win_ntp.py
```diff
@@ -0,0 +1,109 @@
...
+    type: int
+    choices: [ 0, 1, 2 ]
+  sync_clock:
+    description:
+      - It resynchronize its clock as soon as possible.
+      - To update the ntp server with new peerlist.
+    elements: str
+    type: list
+    default: time.windows.com,0x9
```
jborean93, line 56: Can you explain this format any more, what's the `0x9` representing?

jborean93 (commented):
Just did a quick pass through and added some comments. One part that does worry me a bit is how it calls external binaries and doesn't report the stderr in the error message if it fails. This info could be quite helpful for end users trying to debug why things have changed.

plugins/modules/win_ntp.ps1
sakar97: I was thinking, timezone functionality should also be present in this module too, I am aware of the `win_timezone` module but it's easier for users to set time while configuring the NTP.

plugins/modules/win_ntp.py
sakar97: Value “0x9” means – it use special interval + client mode association to the external time source. There are 0x1,0x2,0x4 and 0x8 as SpecialInterval,UseAsFallbackOnly,SymmetricActive,Client Mode flag respectively.

plugins/modules/win_ntp.ps1
```diff
@@ -0,0 +1,189 @@
...
+#AnsibleRequires -CSharpUtil Ansible.Basic
+
+$spec = @{
+  options = @{
+      peerlist = @{ type= "list"; elements = "str" }
+      enabled = @{ type= "bool"; default= "True" }
+      time_zone = @{ type= "str"; default= "UTC" }
+      factory_default = @{ type= "bool"; default= "False" }
+      type = @{ type= "str"; choices = "NTP","NT5DS","NoSync","AllSync"; }
+      cross_site_sync_flags = @{ type= "str"; choices = "0","1","2" }
```
jborean93, line 15: Have the flags be human readable. I shouldn't have to look at the docs to figure out what `cross_site_sync_flags: 0` means. The module can convert our human readable value to the relevant integer and not have the user worry about it at all. See https://github.com/ansible-collections/community.windows/pull/150/files#r515439307.

plugins/modules/win_ntp.py
jborean93: So what we should have is to make this option `type: list` with `elements: dict`. Then those suboptions would be something like `source` and `flags` (or whatever relevant name it is`. Having end users deal with raw integers or bit flags is not ideal. Someone who is reading the tasks should have a generic idea what each parameter means without having to look at the docs. The module can easily parse those string values back to the relevant integer to deal with the getting and setting aspect internally.

maximveksler:
Would be happy for this to get fixed and then merged. Turning on NTP on windows and then doing a "sync now" is a great addition to the automation arsenal.

mcflux:
Whats missing here to make this feature usable?

sakar97:
I was stuck with different tasks. I will try to make changes as per review suggested so this module can be used soon by everyone. Thanks

sgruber94:
@sakar97  did you need help on this?

jborean93:
Closing due to the age of the PR. If someone wishes to continue with the work they can submit a new PR with the changes requested added in.

agibson2:
I was excited when I found this work being done.  Anyone know if this work got into another module I am not aware of?  An idempotent play for this would really be great instead of using regular shell play to just force it every time.  It even looks like sync flags were thought of which is important.

krousenick:
Bump. I really would like to see win_ntp module added. I use this collection quite often. Some one did the work, IMHO I'd rather see less used modules removed to make room for this.

GHDEV00:
Hi @jborean93 

I understand that the collection currently doesn't accept new modules currently. Nevertheless is NTP a central service for any Windows host in a corporate network. Because of this my search for a NTP role also started with looking at the Windows community collection. I assume I am not alone with this when looking at the comments of the pull requests https://github.com/ansible-collections/community.windows/pull/150#top and https://github.com/ansible-collections/community.windows/pull/593 . This leaves the question what must change for a new module to be accepted to the collection?

PR #155, opened 2020-10-26
add computer name parameter to win_dns_zone module

cjpit:
##### SUMMARY
Adds computer name parameter, similar to win_dns_record

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
win_dns_zone

##### ADDITIONAL INFORMATION
https://docs.microsoft.com/en-us/powershell/module/dnsserver/add-dnsserverprimaryzone?view=win10-ps

### DIFF
--- a/plugins/modules/win_dns_record.ps1
+++ b/plugins/modules/win_dns_record.ps1
@@ -16,7 +16,7 @@ $spec = @{
         value = @{ type = "list"; elements = "str"; default = @() ; aliases=@( 'values' )}
         weight = @{ type = "int"}
         zone = @{ type = "str"; required = $true }
-        computer_name = @{ type = "str" }
+        server = @{ type = "str" }
     }
     required_if = @(,@("type", "SRV", @("port", "priority", "weight")))
     supports_check_mode = $true
@@ -33,12 +33,12 @@ $type = $module.Params.type
 $values = $module.Params.value
 $weight = $module.Params.weight
 $zone = $module.Params.zone
-$dns_computer_name = $module.Params.computer_name
+$server = $module.Params.server
 
 
 $extra_args = @{}
-if ($null -ne $dns_computer_name) {
-    $extra_args.ComputerName = $dns_computer_name
+if ($null -ne $server) {
+    $extra_args.ComputerName = $server
 }
 
 if ($state -eq 'present') {
--- a/plugins/modules/win_dns_record.py
+++ b/plugins/modules/win_dns_record.py
@@ -77,12 +77,13 @@ options:
     - The zone must already exist.
     required: yes
     type: str
-  computer_name:
+  server:
     description:
       - Specifies a DNS server.
       - You can specify an IP address or any value that resolves to an IP
         address, such as a fully qualified domain name (FQDN), host name, or
         NETBIOS name.
+    version_added: 1.2.0
     type: str
 '''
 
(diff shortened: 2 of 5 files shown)


plugins/modules/win_dns_zone.py
```diff
@@ -80,6 +80,13 @@
       - At least one server is required.
     elements: str
     type: list
+  computer_name:
```
jborean93, line 83: I'm torn as to whether this should be `server` or `computer_name`. I know the underlying cmdlets use `-ComputerName` but we've already set a precedent for `server` with the AD modules to explicitly select the AD host to target when managing the object.

I personally lean towards making this `server` to make it uniform with what we already do but interested to hear what you think.

plugins/modules/win_dns_zone.py
```diff
@@ -80,6 +80,13 @@
...
+    description:
+      - Specifies a DNS server.
+      - You can specify an IP address or any value that resolves to an IP
+        address, such as a fully qualified domain name (FQDN), host name, or
+        NETBIOS name.
+    type: str
```
jborean93, line 89: This is a new option, it should have `version_added: 1.2.0` to tell users when this was first available in the docs.

jborean93 (changes requested):
Changes look good, just wondering whether we should change the option to `server` to match the way things are done in the AD modules. 

The only other thing we want to add is a changelog fragment so we can advertise this new feature when we release the next version of this collection. To add a fragment you need to create a new file that ends with `.yml` in https://github.com/ansible-collections/community.windows/tree/main/changelogs/fragments.

The contents would be something like (feel free to change the text):

```yaml
minor_changes:
- win_dns_zone - Added ``module option`` to specify the DNS server to connect to when managing the zone object
```

plugins/modules/win_dns_zone.py
cjpit: Yeah I'd agree, server is easier to understand - When I first read the docs for the module, i didn't get it and it wasn't until i got the powershell cmd working I realized what to do, so yep can make those changes, as well as to win_dns_record

plugins/modules/win_dns_record.py
```diff
@@ -77,12 +77,13 @@
...
     required: yes
     type: str
-  computer_name:
+  server:
     description:
       - Specifies a DNS server.
       - You can specify an IP address or any value that resolves to an IP
         address, such as a fully qualified domain name (FQDN), host name, or
         NETBIOS name.
+    version_added: 1.2.0
```
jborean93, line 86: Sorry I was not aware that `computer_name` was already in use here, we still need to accept as an alias that for backwards compatibility. We can skip the `version_added` as technically the option isn't new, just make sure the description mentions it was renamed in the 1.2.0 release like so

```suggestion
  server:
    description:
      - Specifies a DNS server.
      - You can specify an IP address or any value that resolves to an IP
        address, such as a fully qualified domain name (FQDN), host name, or
        NETBIOS name.
      - Was renamed from I(computer_name) to I(server) in 1.2.0.
    aliases:
    - computer_name
```

plugins/modules/win_dns_record.ps1
```diff
@@ -16,7 +16,7 @@ $spec = @{
         value = @{ type = "list"; elements = "str"; default = @() ; aliases=@( 'values' )}
         weight = @{ type = "int"}
         zone = @{ type = "str"; required = $true }
-        computer_name = @{ type = "str" }
+        server = @{ type = "str" }
```
jborean93, line 19: Need to make sure we don't break any existing playbooks by accepting it as an alias

```suggestion
        server = @{ type = "str"; aliases = @("computer_name") }
```

PR #158, opened 2020-11-05
Feature/nssm appenv and custom path

morco:
##### SUMMARY

This extends `win_nssm` with two parameters which can be used to populate the nssm field `AppEnvironmentExtra` internally: 

- `app_environment`: generic key value pairs passed through to service app environment
- `custom_pathvar`: used for setting the $PATH envvar for service app, it makes it easy to merge the default PATH with custom appends and / or prepends

**IMPORTANT:** For specific situations this PR might result in breaking changes. If someone uses the win_nssm module but handles filling `AppEnvironmentExtra` by different means (for example manually or with a custom `win_shell` call), the app environment would be reset to empty after the win_nssm call even when optional new param `app_environment` is unset. Not sure if this is a real usecase and if it might be better to change this, so that only if `app_environment` is explicitly added to a task it is active, but conceptionally this seems imho the right think to do, because if something is managed by an ansible module, it should be managed completly by it, right? 
And if the task in the playbook does not define any app environment the actual service should also not have one.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
win_nssm

##### ADDITIONAL INFORMATION

```paste below

```


jborean93:
I've yet to look at the actual code but I'm wondering whether we need to make `custom_pathvar` as complex as it is right now. why not just have `app_environment` take in the `PATH` environment var and have it include whatever the user desires. If they wish to append or prepend it to the existing `PATH` env var on the host they can use the `{{ ansible_env.PATH }}` fact or get the PATH value manually beforehand with `win_shell`. This keeps things simple where env vars are all defined under 1 key and make it completely explicit as to the `PATH` env var the user wishes to define for their service.

morco:
Yeah, you probably right about `custom_pathvar`. Accomplishing the same just using plain ansible vars, filters and such is actually really easy, so making a builtin special case for this seems over the top, so I removed it. I also added an example for `app_environment`. What about the "breaking" behaviour for `app_environment`: 

   Omit == Explicity set to null == Explicity set to empty dict => Ansible forces nssm app_env to empty

Is it ok to do this because it is architectorial the right thing (it is, isn't it?) or would it be better to play safe here and map the omit case to "dont mess with app_env at all" ? As argued above it would only break things for playbooks which use some kind of combination of a custom homebrew mechanism to set `app_enviroment` together with standard `win_nssm` module.
Than again I'm really not sure if I can actually distinguish inside module code between user ommitted an option and user explicitly set it to null / empty.

I could describe the situation in changelog `know_issues` section, maybe that would be the best thing todo? Or even classify it as `breaking_change`, but if limiting the focus on just the module itself nothing is breaking.

jborean93:
> Is it ok to do this because it is architectorial the right thing (it is, isn't it?) or would it be better to play safe here and map the omit case to "dont mess with app_env at all" 

I would say the safe thing is when `app_environment` is not defined it should work with the current behaviour and not touch `AppEnvironmentExtra` at all. While you could argue that we should always manage it I don't think it's worth making a breaking change in behaviour right off the bat.

What we could do is not touch when it isn't defined, check to see if it is set externally, and add a deprecation warning saying in the future the module will set this to `@{}` by default. That should really be in a separate PR if you wanting to go down that route though.

> Than again I'm really not sure if I can actually distinguish inside module code between user ommitted an option and user explicitly set it to null / empty.

You would have to remove `-default @{}` in `Get-AnsibleParam` and thus if it is omitted by the user then it will become `$null`. Now you can differentiate between:

* `$null` - do nothing
* `@{}` - user explicitly passed in an empty dict
* `@{Key = 'value'}` - user explicitly passed in dict of env vars

plugins/modules/win_nssm.ps1
```diff
@@ -477,6 +483,16 @@ if ($state -eq 'absent') {
         Update-NssmServiceParameter -parameter "AppStdout" -value $stdoutFile @common_params
         Update-NssmServiceParameter -parameter "AppStderr" -value $stderrFile @common_params
 
+        # note: convert app_env dictionary to list of strings in the form key=value and pass that a long as value
+        $tmp = $( $app_env.GetEnumerator() | ForEach-Object { "$($_.Name)=$($_.Value)" } )
```
jborean93, line 487: I would use a proper name for this var like `$app_env_strings` or something. The subexpression isn't doing anything special here as well so you may as well make it like

```suggestion
        $tmp = $app_env.GetEnumerator() | ForEach-Object { "$($_.Name)=$($_.Value)" }
```

plugins/modules/win_nssm.ps1
```diff
@@ -38,6 +38,12 @@ $stderrFile = Get-AnsibleParam -obj $params -name "stderr_file" -type "path"
 
 $executable = Get-AnsibleParam -obj $params -name "executable" -type "path" -default "nssm.exe"
 
+$app_env = Get-AnsibleParam -obj $params -name "app_environment" -type "dict" -default @{}
```
jborean93, line 41: If you want to determine if this was set or not change it to the following

```suggestion
$app_env = Get-AnsibleParam -obj $params -name "app_environment" -type "dict"
```

If not defined then it will be `$null`. Also just an FYI the `-type "dict"` does nothing for the legacy parameter handling. Unfortunately until this module is swapped to the `Ansible.Basic` style there's not much we can do about that and just expect the user to input the data properly for now.

plugins/modules/win_nssm.py
```diff
@@ -100,6 +100,10 @@
     type: str
     choices: [ auto, delayed, disabled, manual ]
     default: auto
+  app_environment:
+    description:
+      - Key/Value pairs which will be added to the environment of the service application.
+    type: dict
```
jborean93, line 106: Should have `version_added: 1.2.0` to denote this was added in that collection version.

plugins/modules/win_nssm.ps1
morco: Done

plugins/modules/win_nssm.ps1
morco: Done

plugins/modules/win_nssm.py
morco: Done

morco:
Ok,
I have handled the review points, changed the behaviour so that it does nothing when 
app_env is null, and added a changelog fragment, so I think I'm done now? Let me know 
if I missed something.

jborean93 (approved):
Awesome, the changes are looking great. Thanks for working on this and a big thanks for writing the tests.

PR #159, opened 2020-11-11
win_partition: Return size as number when set in bytes

silid:
##### SUMMARY
Small change to ensure size is returned as a number when set in bytes

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
win_partition

##### ADDITIONAL INFORMATION
win_partition can return incorrect size comparisons when the size is submitted in bytes as the value was returned as a string rather than a number.

This PR returns the partition size as a number even when submitted in bytes so the size comparison works.

Before change:
```paste below
    "msg": "Specified partition size (53687091200) exceeds size supported by the partition (107238899200)"

```

### DIFF
--- a/changelogs/fragments/159-win_partition_bytes_as_number.yaml
+++ b/changelogs/fragments/159-win_partition_bytes_as_number.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+  - win_partition - fix size comparison errors when size specified in bytes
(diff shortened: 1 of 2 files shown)


briantist (approved):
This looks good to me! Maintainers might ask for other changes though.

You'll need a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to) as well.

changelogs/fragments/159-win_partition_bytes_as_number.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - win_partition - fix size comparison errors when size specified in bytes
```
briantist, line 2:
```suggestion
  - win_partition - fix size comparison errors when size specified in bytes (https://github.com/ansible-collections/community.windows/pull/159)
```

jborean93:
Seems good to me, thanks for fixing this problem.

PR #170, opened 2020-12-01
win_psrepository_copy - new module

briantist:
##### SUMMARY

Adds new module `win_psrepository_copy` for copying PSRepositories to other user profiles on the system.
This solves for the lack of host-wide registered PSRepositories.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

win_psrepository_copy

### DIFF
--- a/plugins/modules/win_psrepository_copy.ps1
+++ b/plugins/modules/win_psrepository_copy.ps1
@@ -1,0 +1,234 @@
+#!powershell
+
+# Copyright: (c) 2020, Brian Scholer <@briantist>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+#AnsibleRequires -CSharpUtil Ansible.Basic
+
+$spec = @{
+    supports_check_mode = $true
+    options = @{
+        source = @{
+            type = 'path'
+            default = '%LOCALAPPDATA%\Microsoft\Windows\PowerShell\PowerShellGet\PSRepositories.xml'
+        }
+        name = @{
+            type = 'list'
+            elements = 'str'
+            default = @(
+                '*'
+            )
+        }
+        exclude = @{
+            type = 'list'
+            elements = 'str'
+        }
+        profiles = @{
+            type = 'list'
+            elements = 'str'
+            default = @(
+                '*'
+            )
+        }
+        exclude_profiles = @{
+            type = 'list'
+            elements = 'str'
+            default = @(
+                'systemprofile'
+                'LocalService'
+                'NetworkService'
+            )
+        }
+    }
+}
+
+$module = [Ansible.Basic.AnsibleModule]::Create($args, $spec)
+
+$module.Diff.changed = @{}
+$module.Diff.unchanged = @{}
+
+function Select-Wildcard {
+    <#
+        .SYNOPSIS
+        Compares a value to an Include and Exclude list of wildcards,
+        returning the input object if a match is found
+
+        .DESCRIPTION
+        If $Property is specified, that property of the input object is
+        compared rather than the object itself, but the original object
+        is returned, not the property.
+    #>
+    [CmdletBinding()]
+    [OutputType([object])]
+    param(
+        [Parameter(Mandatory=$true, ValueFromPipeline=$true)]
+        [object]
+        $InputObject ,
+
+        [Parameter()]
+        [String]
+        $Property ,
+
+        [Parameter()]
+        [String[]]
+        $Include ,
+
+        [Parameter()]
+        [String[]]
+        $Exclude
+    )
+
+    Process {
+        $o = if ($Property) {
+            $InputObject.($Property)
+        }
+        else {
+            $InputObject
+        }
+
+        foreach ($inc in $Include) {
+            $imatch = $o -like $inc
+            if ($imatch) {
+                break
+            }
+        }
+
+        if (-not $imatch) {
+            return
+        }
+
+        foreach ($exc in $Exclude) {
+            if ($o -like $exc) {
+                return
+            }
+        }
+
+        $InputObject
+    }
+}
+
+function Get-ProfileDirectory {
+    <#
+        .SYNOPSIS
+        Returns DirectoryInfo objects for each profile on the system, as reported by the registry
+
+        .DESCRIPTION
+        The special "Default" profile, used as a template for newly created users, is explicitly
+        added to the list of possible profiles returned. Public is explicitly excluded.
+        Paths reported by the registry that don't exist on the filesystem are silently skipped.
+    #>
+    [CmdletBinding()]
+    [OutputType([System.IO.DirectoryInfo])]
+    param(
+        [Parameter()]
+        [String[]]
+        $Include ,
+
+        [Parameter()]
+        [String[]]
+        $Exclude
+    )
+
+    $regPL = 'HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion\ProfileList'
+
+    # note: this is a key named "Default", not the (Default) key
+    $default = Get-ItemProperty -LiteralPath $regPL | Select-Object -ExpandProperty Default
+
+    # "ProfileImagePath" is always the local side of the profile, even if roaming profiles are used
+    # This is what we want, because PSRepositories are stored in AppData/Local and don't roam
+    $profiles = (
+        @($default) +
+        (Get-ChildItem -LiteralPath $regPL | Get-ItemProperty | Select-Object -ExpandProperty ProfileImagePath)
+    ) -as [System.IO.DirectoryInfo[]]
+
+    $profiles |
+        Where-Object -Property Exists -EQ $true |
+        Select-Wildcard -Property Name -Include $Include -Exclude $Exclude
+}
+
+function Compare-Hashtable {
+    <#
+        .SYNOPSIS
+        Attempts to naively compare two hashtables by serializing them and string comparing the serialized versions
+    #>
+    [CmdletBinding()]
+    [OutputType([bool])]
+    param(
+        [Parameter(Mandatory)]
+        [hashtable]
+        $ReferenceObject ,
+
+        [Parameter(Mandatory)]
+        [hashtable]
+        $DifferenceObject ,
+
+        [Parameter()]
+        [int]
+        $Depth
+    )
+
+    if ($PSBoundParameters.ContainsKey('Depth')) {
+        $sRef = [System.Management.Automation.PSSerializer]::Serialize($ReferenceObject, $Depth)
+        $sDif = [System.Management.Automation.PSSerializer]::Serialize($DifferenceObject, $Depth)
+    }
+    else {
+        $sRef = [System.Management.Automation.PSSerializer]::Serialize($ReferenceObject)
+        $sDif = [System.Management.Automation.PSSerializer]::Serialize($DifferenceObject)
+    }
+
+    $sRef -ceq $sDif
+}
+
+# load the repositories from the source file
+try {
+    $src = $module.Params.source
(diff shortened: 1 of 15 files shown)


plugins/modules/win_psrepository_copy.ps1
```diff
@@ -0,0 +1,234 @@
...
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+#AnsibleRequires -CSharpUtil Ansible.Basic
+
+$spec = @{
+    supports_check_mode = $true
+    options = @{
+        source = @{
+            type = 'path'
+            default = '%LOCALAPPDATA%\Microsoft\Windows\PowerShell\PowerShellGet\PSRepositories.xml'
```
jborean93, line 13: Curious, does this path ever change. Just wondering why someone would want to explicitly set this? I would assume it's if you want to copy it from another user's local app data dir?

plugins/modules/win_psrepository_copy.ps1
```diff
@@ -0,0 +1,234 @@
...
+        $Include ,
+
+        [Parameter()]
+        [String[]]
+        $Exclude
+    )
+
+    $regPL = 'HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion\ProfileList'
+
+    # note: this is a key named "Default", not the (Default) key
```
jborean93, line 134: Sigh, don't you just love the registry sometimes...

plugins/modules/win_psrepository_copy.ps1
```diff
@@ -0,0 +1,234 @@
...
+        [int]
+        $Depth
+    )
+
+    if ($PSBoundParameters.ContainsKey('Depth')) {
+        $sRef = [System.Management.Automation.PSSerializer]::Serialize($ReferenceObject, $Depth)
+        $sDif = [System.Management.Automation.PSSerializer]::Serialize($DifferenceObject, $Depth)
+    }
+    else {
+        $sRef = [System.Management.Automation.PSSerializer]::Serialize($ReferenceObject)
```
jborean93, line 175: Thank you for showing me this, I always just used Export-Clixml read the file then deleted it which is not ideal.

plugins/modules/win_psrepository_copy.ps1
```diff
@@ -0,0 +1,234 @@
...
+# load the repositories from the source file
+try {
+    $src = $module.Params.source
+    $src_repos = Import-Clixml -LiteralPath $src -ErrorAction Stop
+}
+catch [System.IO.FileNotFoundException] {
+    $module.FailJson("The source file '$src' was not found.", $_)
+}
+catch {
+    $module.FailJson("There was an error loading the source file '$src'.", $_)
```
jborean93, line 191: I still think it's good to include the exception message in the FailJson message.

```suggestion
    $module.FailJson("There was an error loading the source file '$src': $($_.Exception.Message).", $_)
```

plugins/modules/win_psrepository_copy.ps1
```diff
@@ -0,0 +1,234 @@
...
+                'LocalService'
+                'NetworkService'
+            )
+        }
+    }
+}
+
+$module = [Ansible.Basic.AnsibleModule]::Create($args, $spec)
+
+$module.Diff.changed = @{}
```
jborean93, line 47: I've always used `before`/`after` for the diff, does `changed` and `unchanged` also work?

plugins/modules/win_psrepository_copy.py
```diff
@@ -0,0 +1,140 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2020, Brian Scholer <@briantist>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+DOCUMENTATION = r'''
+---
+module: win_psrepository_copy
```
jborean93, line 9: You want to have `version_added: 1.3.0` to indicate when this module was added to the collection.

plugins/modules/win_psrepository_copy.ps1
```diff
@@ -0,0 +1,234 @@
...
+$profiles = Get-ProfileDirectory -Include $module.Params.profiles -Exclude $module.Params.exclude_profiles
+
+foreach ($user in $profiles) {
+    $username = $user.Name
+
+    $repo_dir = $user.FullName | Join-Path -ChildPath 'AppData\Local\Microsoft\Windows\PowerShell\PowerShellGet'
+    $repo_path = $repo_dir | Join-Path -ChildPath 'PSRepositories.xml'
+
+    if (Test-Path -LiteralPath $repo_path) {
+        $cur_repos = Import-Clixml -LiteralPath $repo_path
```
jborean93, line 203: I'm curious, can the repository clixml contain a credential/secure string if one is assigned to that repository. If so that might cause issues trying to deserialize the object as it would be encrypted using that user's credentials.

jborean93 (approved):
Looks great, most of my comments are just clarifications of behaviour for my own sake. The only outstanding changes are really just `version_added` in the docs and a slight tweak to an error message.

plugins/modules/win_psrepository_copy.ps1
briantist: The path where `PowerShellGet` reads it from never changes as far I can tell. The main reason to explicitly set this is if you aren't trying to use the running user (or any existing user's) file as a source (they may have changed it after all), like if you wanted to maintain a known source set in a specific location (or copy it in on execution from Ansible to a temporary location, like I did in the tests).

plugins/modules/win_psrepository_copy.ps1
briantist: Ah yeah, super useful. Occasionally combine it with `Set-Clipboard` / `scb` (and then `gcb` with the `Deserialze()`) to move large complex objects around in debugging, like in and out of RDP sessions, without dealing with a file.

plugins/modules/win_psrepository_copy.ps1
briantist: Yeah that's a good point, I keep thinking that passing in the exception object is enough, but I guess those details will only be shown with high enough verbosity; it's good to have at least something in the main message shown.

plugins/modules/win_psrepository_copy.ps1
briantist: This was actually hard for me to tell.. there's not a lot of documentation about diff output. From what I could tell looking through a few other modules it seems like there's no actual defined structure/rules for what goes into `.Diff`, but I don't see the diff output unless both `--diff` and verbose are used. Is there supposed to be a different behavior and the reason is that I used the wrong fields? 

Happy to be wrong and update this if there's a better way.

plugins/modules/win_psrepository_copy.py
briantist: haha, how easily I catch this on someone _else's_ PR.. will add!

plugins/modules/win_psrepository_copy.ps1
briantist: woah that's a really good point.. You can specify a `Credential` (and `ProxyCredential`), and I can't actually imagine them handling it any other way than the default DPAPI-based encryption. Those details then wouldn't really be copy-able; I'm going to verify and at a minimum probably update the documentation. 

If it does work this way, there's not a good workaround I can even see. In a limited circumstance, if Ansible had the credentials for both the source user and the destination user, it could be possible to do this in 2 steps, but that'd be complicated.

I'll also dig into some way I can import the XML without it failing outright, that way it doesn't affect the copying of unrelated repositories with no credentials.. unfortunately the only way I can think of off the top of my head is modifying the XML before deserialization; I tried hard to avoid that, but I'll see what I can find out.

Thanks for this!

plugins/modules/win_psrepository_copy.ps1
jborean93: Thanks for the explanation.

plugins/modules/win_psrepository_copy.ps1
jborean93: Yea I don't know of a way to overcome this, the closest thing would be to edit out the credentials and copy it without them but I don't think we need to handle that straight away. If anything as you said it's probably just something we need to document and just live with it for now.

briantist: > the closest thing would be to edit out the credentials and copy it without them

Yeah I was thinking about that too.. but f the credentials were needed, I suppose it wouldn't be that helpful.
In any case I'll verify and ensure it's at least explained in the docs/notes. Thanks again, great callout.

plugins/modules/win_psrepository_copy.ps1
jborean93: Yea it's not documented too well unfortunately. When implementing diff support in the past I've used 3 different keys

* before: String/dict/list containing the pre state
* after: String/dict/list containing the post state
* prepared: A string containing your own diff string

If you use before/after then Ansible itself handles the color/formatting itself which is great. An example would be `win_regedit`

image

You can see that Ansible adds the before/after difference similar to how git diff works. It even handled additions/removals to dicts/lists in a very nice way. Ultimately when I add diff support I have the before/after be an empty string to represent absent or a dict with each key being the relevant option for that object at that stage.

The prepared one is something I have used in the past but that was before I knew about before/after. It seems like `win_copy` is still using that. The downside of `prepared` is that the colors are now gone and you need to handle the before/after checks yourself. The plus side is you have full control over how the diff actually looks.

image

When looking at the callback code that generates the diff text it does not seem like `changed`/`unchanged` is treated in any special way compared to `before`/`after` https://github.com/ansible/ansible/blob/484e4af4d0b0f8ba88f73c20f592bb107eb9396a/lib/ansible/plugins/callback/__init__.py#L168. I would recommend using before/after here instead but try it out and see what happens.

plugins/modules/win_psrepository_copy.ps1
briantist: So beyond my expectations... they DO NOT store credentials in the XML file at all. It seems they use separate credential providers for this purpose, probably at least in part to support non-Windows: https://github.com/PowerShell/PowerShellGetv2/blob/f45bfe1a66df7063aa8de10fc4f147fdc23639af/src/PowerShellGet/public/psgetfunctions/Get-CredsFromCredentialProvider.ps1

This means that this module's operations shouldn't be affected, but I'm still unsure if such a copied repository would be usable by another user. It seems to be stored in a cred provider by sourcelocation, so it could be found, but I'm not sure that cross user the creds could actually be retrieved (it may also vary depending on the methods used in that function), and I don't have a working authenticated repository to check.

So probably still worth of a doc update, but likely no code updates; and to add on to that we won't be able to even tell that a given repository requires credentials in order to issue a warning (even `Get-PSRepository` doesn't show that the repo requires creds).

Will make some updates soon.

briantist: Just added to the notes. Not much can be done programmatically to detect/defend against this, and same is true for both proxy params (not included in the XML file).

plugins/modules/win_psrepository_copy.ps1
briantist: Switched to `before`/`after`, seems to work just fine!

jborean93 (approved):

jborean93:
Thanks @briantist for working on this, it's a nice addition to the collection.

plugins/modules/win_psrepository_copy.py
```diff
@@ -8,6 +8,7 @@
 ---
 module: win_psrepository_copy
 short_description: Copies registered PSRepositories to other user profiles
+version_added: '1.3.0'
```
briantist, line 11: omg I forgot it again, thanks @jborean93

PR #180, opened 2020-12-17
Fix copy-paste-error in win_firewall_rule

tortenairbag:
##### SUMMARY

Fixes an copy-paste-error for the service parameter in win_firewall_rule.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

- win_firewall_rule

##### ADDITIONAL INFORMATION

If the task gets executed with "service=any", a firewall rule gets created which effects services named "any" instead of all services.

### DIFF
--- a/changelogs/fragments/win_firewall_rule-service-any.yml
+++ b/changelogs/fragments/win_firewall_rule-service-any.yml
@@ -1,0 +1,2 @@
+bugfixes:
+- 'win_firewall_rule - Ensure ``service: any`` is set to match any service instead of the literal service called ``any`` as per the docs'
--- a/plugins/modules/win_firewall_rule.ps1
+++ b/plugins/modules/win_firewall_rule.ps1
@@ -169,7 +169,7 @@ try {
     if ($null -ne $description) { $new_rule.Description = $description }
     if ($null -ne $group) { $new_rule.Grouping = $group }
     if ($null -ne $program -and $program -ne "any") { $new_rule.ApplicationName = [System.Environment]::ExpandEnvironmentVariables($program) }
-    if ($null -ne $service -and $program -ne "any") { $new_rule.ServiceName = $service }
+    if ($null -ne $service -and $service -ne "any") { $new_rule.ServiceName = $service }
     if ($null -ne $protocol -and $protocol -ne "any") { $new_rule.Protocol = ConvertTo-ProtocolType -protocol $protocol }
     if ($null -ne $localport -and $localport -ne "any") { $new_rule.LocalPorts = $localport }
     if ($null -ne $remoteport -and $remoteport -ne "any") { $new_rule.RemotePorts = $remoteport }


briantist:
Without actually testing it the change itself looks good to me; I'm not sure if other changes (tests, etc.) will be required. At a minimum you'll need to add a [changelog fragment](https://github.com/ansible-collections/community.windows/tree/main/changelogs/fragments).

tortenairbag:
I added the changelog fragment as requested and also included some tests to verify, that the documented default values are working as intended.
The test succeeds with the old version too, because the error occurs if "program" and "service" are set, "service" set to "any" but "program" not.
So the test is not verifying that my bugfix works, but I guess its still nice to have a verification of default parameters.

Writing the tests I recognized, that there was a small error in the documentation - the default value for the parameter "enabled" is "yes" instead of "true", so i changed that too.
And editing the doc file lead me to rewrite it a bit, in the new version all default values are displayed in the "Choices/Defaults" column instead in the comments.

This also implies that the doc file (/docs/community.windows.win_firewall_rule_module.rst) needs to be changed too, but I assume that the .rst docs are rebuilt on each release (history of /docs folder: https://github.com/ansible-collections/community.windows/commits/main/docs)
Am I right or are doc changes also in scope of pull requests?

changelogs/fragments/180-win_firewall_rule-copy-paste-error.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+- win_firewall_rule - Fixed default value handling of service parameter
```
briantist, line 2:
```suggestion
- win_firewall_rule - fixed default value handling of ``service`` parameter (https://github.com/ansible-collections/community.windows/pull/180).
```

briantist:
> I added the changelog fragment as requested and also included some tests to verify, that the documented default values are working as intended.
> The test succeeds with the old version too, because the error occurs if "program" and "service" are set, "service" set to "any" but "program" not.
> So the test is not verifying that my bugfix works, but I guess its still nice to have a verification of default parameters.
> 

It sounds like that new test may be testing for something useful, but a test that catches the bug you're fixing would be best to catch future regressions.

> Writing the tests I recognized, that there was a small error in the documentation - the default value for the parameter "enabled" is "yes" instead of "true", so i changed that too.

`yes`/`no` are valid forms of specifying booleans so they are equivalent. Typically the form used when writing values in documentation, examples, and playbooks themselves is whichever sounds the most natural.

[This is a YAMLism, not specific to Ansible itself](https://stackoverflow.com/a/47877502/3905079).

> And editing the doc file lead me to rewrite it a bit, in the new version all default values are displayed in the "Choices/Defaults" column instead in the comments.
> 

Someone more familiar with how this module works will have to comment, but I think you may need to revert some or all of these changes.

One thing I noticed was that the values described in the description specifically said it was a default for _new_ rules. To me this implies that when you specify an existing rule (when a rule would be modified and not created new), its value for that particular parameter will not be changed.

If you set the default in the arg spec, the issue becomes that the module then cannot determine whether the value was explicitly supplied by the user, or whether it was just given as a "default" value. That would cause the module to overwrite that parameter on any rule that doesn't match the default unless the caller explicitly knew the value and passed it in.

Again I'll stress I haven't looked closely at the code and this is speculation, but do be careful; it might be best to look at these changes separately and in a new PR.

I'll also say that if what I wrote above is true, then it is concerning that the tests passed, as it means the tests possibly don't account for that (but that's not on you!).

> This also implies that the doc file (/docs/community.windows.win_firewall_rule_module.rst) needs to be changed too, but I assume that the .rst docs are rebuilt on each release (history of /docs folder: https://github.com/ansible-collections/community.windows/commits/main/docs)
> Am I right or are doc changes also in scope of pull requests?

The RST docs are generated based on the module docs, so updates there are all that is needed.

jborean93 (approved):
Thanks for the fix, I've just added a change which includes the changelog fragment so users can understand what changes were made in each collection version. While @briantist is right that we would like to see integration tests accompanied bugfixes and features I think for this rudimentary problem it's ok to exclude it.

Tests are great to ensure that future changes don't re-introduce a bug as now we have a test for that behaviour.

PR #187, opened 2021-01-24
Remove Shippable

jborean93:
##### SUMMARY
We no longer need to run CI on Shippable.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
all


jborean93:
Looks like I need to fix up win_scoop and a unit test, they can be done in a separate PR though.

PR #193, opened 2021-02-16
WIP: New Module: win_gpo_force

sgruber94:
##### SUMMARY

The win_gpo_force module forces sysvol Folder Replication and updates GroupPolicy

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

win_gpo_force

##### ADDITIONAL INFORMATION


mwtrigg:
Are there any special requirements regarding authentication?  It initiates a remote session on any domain member, and then opens additional remote sessions from the target against all domain controllers, and then a separate remote command against all children of a named OU.  This seems to be potentially fraught with double-hop obstacles, and could lead to numerous failed connections, and failed runs

---

At written (if I read this correctly), it appears that any single failed connection (to a domain controller or client) fails the whole task; I think this is undesirable behavior.

I think it preferable to write a module that performs an operation against a single host, instead of one task that potentially touches thousands of machines - when only run against a single host.  For example, instead of hitting all domain controllers and potentially all computers in a domain, it be structured such that it connects to a single domain controller, optionally triggers replication of directory changes, and optionally triggers replication of DFSR content.  Individual failures could be identified and resolved; and the whole task doesn't fail because one domain controller or client is offline.

Then a second play or separate playbook would include a win_shell task (or a new separate gpupdate module) that operates against individual hosts in the group.  This would allow for per-host status/results.

sgruber94:
You are right.
It will fails the hole task.

I will rewrite / change the module  
*  Force Replication of all Partitions between Domain Controllers / Sysvols
*  gpupdate single client 

Also, I change the input from ou to hostname.

Good Point, documentation misses Authentication / Permissions, will add that also

plugins/modules/win_gpo_force.ps1
```diff
@@ -0,0 +1,75 @@
...
+    }
+    supports_check_mode = $false
+}
+$module = [Ansible.Basic.AnsibleModule]::Create($args, $spec)
+$mode = $module.Params.mode
+$ou = $module.Params.ou
+try {
+    Import-Module GroupPolicy
+} catch {
+    $module.FailJson("win_gpo_force requires the GroupPolicy PS module to be installed")
```
jborean93, line 20: When you use FailJson it takes an optional 2nd argument being either an ErrorRecord or Exception. If you can it's best to supply it so the error reported back to Ansible has more context behind it.

```suggestion
    $module.FailJson("win_gpo_force requires the GroupPolicy PS module to be installed", $_)
```

plugins/modules/win_gpo_force.ps1
```diff
@@ -0,0 +1,75 @@
...
+function Invoke-forcedgpo {
+    param(
+        [string]$oucomputer	# Push gpupdate after AD & SYSVOL Replication
+    )
+    # Check for DFSR cmd tools
+    if ((Get-WindowsFeature -Name RSAT-DFS-Mgmt-Con).InstallState -ne "Installed") {
+        Install-WindowsFeature -Name RSAT-DFS-Mgmt-Con
+    }
+    # Check powershell console for UAC
+    #requires -RunAsAdministrator
```
jborean93, line 31: This won't do anything in Ansible. If you want to make sure the code is running as admin then you will need to check the groups yourself and fail. Personally I think it should just be removed from the code as the majority of the cases will be running as Admin anyway.

plugins/modules/win_gpo_force.ps1
```diff
@@ -0,0 +1,75 @@
...
+    # Force Replication of all Partitions between Domain Controllers
+    (Get-ADDomainController -filter *).Name | ForEach-Object { repadmin /syncall $_ (Get-ADDomain).DistinguishedName /e /A | Out-Null }
```
jborean93, line 33: You use this twice, better off storing the list of AD controllers into a var then doing `$controllers | ForEach-Object ...` in both sections instead of running the query twice.

plugins/modules/win_gpo_force.ps1
```diff
@@ -0,0 +1,75 @@
...
+    Start-Sleep 5
```
jborean93, line 34: Is there anyway to avoid this sleep. Sleeping is never a good idea as

1. It's never guaranteed to wait long enough, what if the host is slower than normal
2. It wastes time if what you are waiting for is already ready, i.e. it's ready in 1 second but the script still waits the additional 4.

plugins/modules/win_gpo_force.ps1
```diff
@@ -0,0 +1,75 @@
...
+    Get-ADReplicationPartnerMetadata -Target "$env:Userdnsdomain" -Scope Domain | Format-Table Server, LastreplicationSuccess -a
```
jborean93, line 35: What's the point of this, it will send the table as output to this function but you aren't capturing it anywhere. For an Ansible module you need to avoid outputting anything as we use the output to return the module result json. Mixing in other objects can cause all sorts of issues so avoid this whenever you can.

plugins/modules/win_gpo_force.ps1
```diff
@@ -0,0 +1,75 @@
...
+    # Sync Sysvol
+    $localdc = $env:computername
+    $remotedc = @() + ((Get-ADDomainController -filter *).Name | Where-Object { $_ -notlike $localdc })
+    # Sync local
+    DFSRDIAG PollAD
+    Start-Sleep 5
+    # Sync remote
+    Foreach ($remote in $remotedc) {
+        $scriptblock = $ExecutionContext.InvokeCommand.NewScriptBlock("DFSRDIAG PollAD")
+        $session = New-PSSession -computername $remote
```
jborean93, line 45: This is not going to work for the majority of authentication setups that Ansible uses. Essentially you will have to be either running with CredSSP, Kerberos with credential delegation, or using become on the tasks.

plugins/modules/win_gpo_force.ps1
```diff
@@ -0,0 +1,75 @@
...
+        $result.changed = $true
+    }
+}
+if($mode -eq "forceupdate") {
+    if ($ou) {
+        try {
+            Invoke-forcedgpo -OuComputer $ou
+            $module.Result.changed = $true
+        } catch {
+            $module.FailJson("an exception occurred when updating GPO - $($_.Exception.Message)")
```
jborean93, line 65: We try to avoid blanket catch blocks in modules. Uncaught exceptions are automatically reported back to the controller and the stack trace shown when you run with `-vvv`. By catching it here you loose all that information making it hard to track down what actually failed.

jborean93 (changes requested):
Apologies for the delay in reviewing this but there are a few concerns I have with this module.

* It seems to be more like do an action rather than make a state change. This is more suited towards using the `script` or `win_shell` module to run the actions you desire
  * I'm not sure you can really do this as an idempotent kind of behaviour at all really
* It does more PSRemoting sessions internally to run commands on each domain controller which will require a credential delegation fix
  * People using NTLM over HTTPS, Kerberos without delegation (large majority of domain users) will never work with this
* The use of sleep is very worrisome, usually it's a sign of a fragile operation that can break when used in different environments

As such I'm not sure if we want to accept something like this in this collection. The way the action is written makes it seem like it would be more suited towards an Ansible role due to the nested PSSession invocations so that might be a better idea going forward.

jborean93:
Unfortunately due to the complexity here I don't see this module ever being merged into this collection. Trying to do nested PSSessions in Ansible is not going to work well. This might be better suited to having it in your own collection so you can iterate and try out new things without being stuck making it backwards compatible or tied to our release cadence.

PR #194, opened 2021-02-16
New Module: win_gpo_link

sgruber94:
### SUMMARY
The wwin_gpo_manage module manage a GroupPolicy

### ISSUE TYPE
- New Module Pull Request

### COMPONENT NAME
win_gpo_manage

### ADDITIONAL INFORMATION


plugins/modules/win_gpo_link.ps1
```diff
@@ -0,0 +1,278 @@
...
+$module = [Ansible.Basic.AnsibleModule]::Create($args, $spec)
+$state = $module.Params.state
+$path = $module.Params.path
+$gponame = $module.Params.gponame
+$gpoorder = $module.Params.order
+$gpodomain = $module.Params.domain
+$gposerver = $module.Params.server
+$enforcedbool = $module.Params.enforced
+$gpolinkenabledbool = $module.Params.linkenabled
+$ErrorActionPreference = 'Stop'#Error Action
```
jborean93, line 172: We automatically do this when invoking a module, it's not really needed here.

plugins/modules/win_gpo_link.ps1
```diff
@@ -0,0 +1,278 @@
...
+try {
+    Import-Module GroupPolicy
+} catch {
+    $module.FailJson("win_grouppolicy requires the GroupPolicy  PS module to be installed")
```
jborean93, line 176: By adding the error record as a 2nd argument to `FailJson` we preserve the error details and sent it back to the controller so it can display it better for users. Also the module name was incorrect.

```suggestion
    $module.FailJson("win_gpo_link requires the GroupPolicy  PS module to be installed", $_)
```

plugins/modules/win_gpo_link.ps1
```diff
@@ -0,0 +1,278 @@
...
+}
+try {
+    Import-Module activedirectory
+} catch {
+    $module.FailJson("win_grouppolicy requires the activedirectory  PS module to be installed")
```
jborean93, line 181: Ditto here

```suggestion
    $module.FailJson("win_gpo_link requires the activedirectory  PS module to be installed", $_)
```

plugins/modules/win_gpo_link.ps1
```diff
@@ -0,0 +1,278 @@
...
+    $enforced = "yes"
+} else {
+    $enforced = "no"
+}
+#States
+if ($state -eq "query") {
+    foreach ($pa in $path) {
+        [string]$path = $pa
+        if ($path) {
+            get-gplink -path $path
```
jborean93, line 200: You aren't doing anything with this result causing it to be sent back to the controller. You should avoid that at all costs. Either `| Out-Null` the result or capture in a variable if you need it for later. Based on the context it sounds like you want to return this information as a structured object. You need to store it into a var and set `$module.Result.key = 'value'` so it's returned back to the controller.

Finally we try to avoid having modules return and set information. `state=query` should be removed from this module and the query functionality should be implemented in a `win_gpo_link_info` module instead.

plugins/modules/win_gpo_link.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2021 Sebastian Gruber ,dacoso GmbH All Rights Reserved.
+# SPDX-License-Identifier: GPL-3.0-only
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+DOCUMENTATION = r'''
+---
+module: win_gpo_link
+author: Sebastian Gruber (@sgruber94)
```
jborean93, line 10: Needs to have `version_added: 1.4.0` to signal when this module was added to `community.windows`.

plugins/modules/win_gpo_link.py
```diff
@@ -0,0 +1,86 @@
...
+short_description: Link a GroupPolicy to a specific ActiveDirectory OU
+description:
+- The C(win_gpo_link) module can link GroupPolicy to a specific ActiveDirectory OU.
+requirements:
+  - PowerShell Module GroupPolicy (https://docs.microsoft.com/en-us/powershell/module/grouppolicy)
+  - PowerShell Module activedirectory (https://docs.microsoft.com/en-us/powershell/module/activedirectory)
+seealso:
+  - module: community.windows.win_gpo_force
+  - module: community.windows.win_gpo_manage
```
jborean93, line 19: These don't exist yet. I would hold off on adding them until the modules themselves are merged. These values are used to generate our docs and if the module does not exist yet then that process will fail.

plugins/modules/win_gpo_link.ps1
```diff
@@ -0,0 +1,278 @@
...
+                #we create an object representing each GPOs, its links status and link order
+                $return = New-Object psobject
+                $return | Add-Member -membertype NoteProperty -Name "Target" -Value $target
+                $return | Add-Member -membertype NoteProperty -Name "GPOID" -Value $gpoguid
+                $return | Add-Member -membertype NoteProperty -Name "DisplayName" -Value $gponame
+                $return | Add-Member -membertype NoteProperty -Name "Domain" -Value $gpodomain
+                $return | Add-Member -membertype NoteProperty -Name "Enforced" -Value $enforced
+                $return | Add-Member -membertype NoteProperty -Name "Enabled" -Value $enabled
+                $return | Add-Member -membertype NoteProperty -Name "Order" -Value $order
+                $return
```
jborean93, line 146: This is quite an old style of creating a PSObject in PowerShell, it's a lot more idiomatic to do

```suggestion
                [PSCustomObject]@{
                    Target = $target
                    GPOID = $gpoguid
                    DisplayName = $gponame
                    Domain = $gpodomain
                    Enforced = $enforced
                    Enabled = $enabled
                    Order = $order
                }
```

jborean93 (changes requested):

plugins/modules/win_gpo_link.ps1
sgruber94: I will generate a new  module for that.

jborean93:
Thanks for your contribution here, unfortunately at this moment I'm not sure we can accept your PRs for new modules. It's not a question of quality but rather whether we have the required time to maintain this work in the `community.windows` repository. A lot of the new modules you are contributing are very domain specific, i.e. for GPOs or DNS/DHCP and in the past we have accepted some of these contributions but going forward that may not be the case.

Instead what I recommend you do is to create your own collection, or collections, that contain your work and distribute that through Galaxy. This allows you to make the changes you desire without being hamstrung by myself in trying to get reviews done and a release created. This also allows you to set the standard/quality level you desire for your own work rather than being stuck with our requirements.

I do appreciate the time and effort you have gone through to submit these PRs. I also apologise for the time it has taken for me to come re-review these submissions but my lack of time for this is a major reason why having your collection is desirable here. If you wish for any help in setting up a collection, or just want to discuss things about Windows you are more than welcome to come talk to us on #ansible-community or #ansible-windows in IRC.

PR #195, opened 2021-02-16
New Module:  win_gpo_manage

sgruber94:
### SUMMARY
The wwin_gpo_manage module manage a GroupPolicy

### ISSUE TYPE
- New Module Pull Request

### COMPONENT NAME
win_gpo_manage

### ADDITIONAL INFORMATION

### DIFF
--- a/plugins/modules/win_gpo_manage.ps1
+++ b/plugins/modules/win_gpo_manage.ps1
@@ -1,0 +1,176 @@
+#!powershell
+# Copyright: (c) 2020 Sebastian Gruber ,dacoso GmbH All Rights Reserved.
+# SPDX-License-Identifier: GPL-3.0-only
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+# AnsibleRequires -CSharpUtil Ansible.Basic
+
+$ErrorActionPreference = 'Stop'#Error Action
+$spec = @{
+    options             = @{
+        name     = @{ type = "list" }
+        folder   = @{ type = "path"; default = "C:\GPO" }
+        mode     = @{ type = "str"; choices = "import", "query", "remove", "export"; default = "import" }
+        override = @{ type = "bool"; default = $false }# Skip if import gpo
+        log_path = @{ type = "str" }
+    }
+    required_if         = @(, 
+        @("mode", "import", @("folder")), 
+        @("mode", "export", @("folder")),
+        @("mode", "remove", @("name"))
+        )
+    supports_check_mode = $true
+}
+$module = [Ansible.Basic.AnsibleModule]::Create($args, $spec)
+$check_mode = $module.CheckMode
+$gpomode = $module.Params.mode
+$removedgpo = $module.Params.name
+$folderpath = $module.Params.folder
+$log_path = $module.Params.log_path
+try {
+    Import-Module GroupPolicy
+} catch {
+    $module.FailJson("win_grouppolicy requires the GroupPolicy PS module to be installed")
+}
+Function Write-DebugLog {
+    Param(
+        [string]$msg
+    )
+    $DebugPreference = "Continue"
+    $ErrorActionPreference = "Continue"
+    $date_str = Get-Date -Format u
+    $msg = "$date_str $msg"
+
+    Write-Debug $msg
+    if($log_path) {
+        Add-Content $log_path $msg
+    }
+}
+function Import-GPOs {
+    param (
+        $importfolder,
+        [bool]$override
+    )
+    $Folder = Get-childItem -Path $Importfolder -Directory
+    $gpos = Get-GPO -All
+    if($folder.count -gt 0) {
+        foreach ($Entry in $Folder) {
+            $override
+            $Name = $Entry.Name + $postname
+            if(($gpos -contains $Name) -and !($override)) {
+                $module.FailJson("GroupPolicy exists already")
+            } else {
+                #GPO does not exist or override is §true
+                Write-DebugLog "Import GPO $name"
+                $Path = $Importfolder + "\" + $entry.Name
+                 (get-item $Path+'\manifest.xml').Attributes += 'Hidden' #make manifest hidden
+                $ID = Get-ChildItem -Path $Path
+                if(!($gpos -contains $Name)) {
+                    #if no GPO exist
+                    New-GPO -Name $Name -WhatIf:$check_mode
+                    $module.result.changed = $true
+                }
+                Import-GPO -TargetName $Name -Path $Path -BackupId $ID.Name -WhatIf:$check_mode
+                $module.result.imported.Add($Name)
+                $module.result.changed = $true
+            }
+        }
+    } else {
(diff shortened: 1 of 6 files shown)


plugins/modules/win_gpo_manage.ps1
```diff
@@ -0,0 +1,147 @@
...
+# Copyright: (c) 2021 Sebastian Gruber ,dacoso GmbH All Rights Reserved.
+# SPDX-License-Identifier: GPL-3.0-only
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+#AnsibleRequires -CSharpUtil Ansible.Basic
+
+$spec = @{
+    options  = @{
+        name     = @{ type = "list"; elements = "str" }
+        folder   = @{ type = "path"; default = "C:\GPO" }
+        mode     = @{ type = "str"; choices = "import", "query", "remove", "export";required=$true }
```
jborean93, line 11: Use `state` instead as it's a more common option in Ansible. It should have the following options

* `import` -> `present`
* `query` -> this shouldn't be here but in a separate `win_gpo_manage_info` module
* `remove` -> `absent`
* `export` -> `exported`

It should also have a default of `present` just like most of the other modules that have `state`.

plugins/modules/win_gpo_manage.ps1
```diff
@@ -0,0 +1,147 @@
...
+}
+$module = [Ansible.Basic.AnsibleModule]::Create($args, $spec)
+$check_mode = $module.CheckMode
+$gpomode = $module.Params.mode
+$removedgpo = $module.Params.name
+$folderpath = $module.Params.folder
+try {
+    Import-Module GroupPolicy
+} catch {
+    $module.FailJson("win_grouppolicy requires the GroupPolicy PS module to be installed")
```
jborean93, line 29:
```suggestion
    $module.FailJson("win_gpo_manage requires the GroupPolicy PS module to be installed", $_)
```

plugins/modules/win_gpo_manage.ps1
```diff
@@ -0,0 +1,147 @@
...
+    }
+    $gpo = Get-GPO -All
+    foreach ($Entry in $gpo) {
+        $Path = $ExportFolder + "\" + $entry.Displayname
+        if((Test-Path -LiteralPath $Path) -and !($override)) {
+            $module.FailJson("Folder already exists, Override is false ")
+        } else {
+            if(!(Test-Path -LiteralPath $Path)) {
+                #Folder doesnt exist
+                New-Item -ItemType directory -Path $Path -WhatIf:$check_mode
```
jborean93, line 79: I'm pretty sure `New-Item` outputs an object, you need to pipe to Out-Null or assign it to a var so it doesn't impact our module output result.

plugins/modules/win_gpo_manage.ps1
```diff
@@ -0,0 +1,147 @@
...
+        }
+    }
+}
+if ($diff_mode) {
+    $module.result.diff = @{}
+}
+if ($gpomode -eq "remove") {
+    #Removes GPOs
+    $gpo = (Get-GPO -All).Displayname
+    $module.diff.before = $gpo | Out-String
```
jborean93, line 96: No need for the `| Out-String`, the diff builder is able to compare a list in before and after and show what was added/removed.

plugins/modules/win_gpo_manage.ps1
```diff
@@ -0,0 +1,147 @@
...
+            if((Test-Path -LiteralPath $Path) -and ($override)) {
+                Remove-Item -ItemType directory -Path $Path -WhatIf:$check_mode
+                New-Item -ItemType directory -Path $Path -WhatIf:$check_mode
+            }
+            Backup-GPO -Guid $Entry.id -Path $Path -WhatIf:$check_mode
+            $module.result.exported.Add($entry.Displayname)
+        }
+    }
+}
+if ($diff_mode) {
```
jborean93, line 90: You haven't defined `$diff_mode`, use `$module.DiffMode` to check if diff mode is enabled. Ultimately you don't even need to check. You can always set the diff settings and the controller is able to selectively display it if `--diff` was enabled. In fact I recommend you always set it so your code runs the diff stuff and won't randomly fail due to a diff specific setting when `--diff` is enabled.

plugins/modules/win_gpo_manage.ps1
```diff
@@ -0,0 +1,147 @@
...
+    $module.result.diff = @{}
```
jborean93, line 91: You should use `$module.Diff` instead to set diff values. It's already pre-populated with an empty hashtable.

plugins/modules/win_gpo_manage.ps1
```diff
@@ -0,0 +1,147 @@
...
+    if ($diff_mode -and $result.changed) {
+        if (!$check_mode) {
+            #only check
+            $module.result.diff.after = [Array]$removedgpo | Out-String
+        } else {
+            #remove gpos
+            $gpoafter = (Get-GPO -All).Displayname
+            $after = [System.Collections.Generic.List`1[String]]
+            $after.AddRange($gpoafter)
+            $module.result.diff.after = $after | Out-String
```
jborean93, line 124: `$after` is now a type and not the list you want it to be so running `$after.AddRange($gpoafter)` will not work. Ultimately you can simplify this to just

```suggestion
            $module.Diff.after = @((Get-Gpo -All).Displayname)`
```

plugins/modules/win_gpo_manage.ps1
```diff
@@ -0,0 +1,147 @@
...
+        }
+    }
+}
+if ($gpomode -eq "query") {
```
jborean93, line 128: This should be in it's own `win_gpo_manage_info` module (really probably `win_gpo_info`.

plugins/modules/win_gpo_manage.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2021 Sebastian Gruber ,dacoso GmbH All Rights Reserved.
+# SPDX-License-Identifier: GPL-3.0-only
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+DOCUMENTATION = r'''
+---
+module: win_gpo_manage
```
jborean93, line 9: Needs `version_added: '1.4.0'`.

jborean93 (changes requested):

plugins/modules/win_gpo_manage.ps1
sgruber94: i will generate a new module for that

briantist:
@sgruber94 it looks like you added a merge commit instead of rebasing your branch against main.. there's a ton of "changes" in here that shouldn't be here. I'm not entirely sure how to fix it, but it seems like you need a rebase, at a minimum.

jborean93:
See https://github.com/ansible-collections/community.windows/pull/194#issuecomment-877442730

PR #199, opened 2021-02-17
Added Authenctiation for win_psmodule and win_psrepository.

nduijvelshoff:
##### SUMMARY
Provides support for #45 by allowing username and password being set to authenticate against a private repository.

##### ISSUE TYPE
- Feature Pull Request

### DIFF
--- a/plugins/modules/win_psmodule.py
+++ b/plugins/modules/win_psmodule.py
@@ -1,6 +1,7 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
 
+# Copyright: (c) 2021, Nick Duijvelshoff <nick@duijvelshoff.com>
 # Copyright: (c) 2018, Wojciech Sciesinski <wojciech[at]sciesinski[dot]net>
 # Copyright: (c) 2017, Daniele Lazzari <lazzari@mailup.com>
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
@@ -59,6 +60,12 @@ options:
     description:
       - Name of the custom repository to use.
     type: str
+  username:
+    description:
+      - Username to authenticate against private repository.
+  password:
+    description:
+      - Password to authenticate against private repository.
   url:
     description:
       - URL of the custom repository to register.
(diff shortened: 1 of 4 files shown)


plugins/modules/win_psmodule.py
```diff
@@ -59,6 +60,12 @@
     description:
       - Name of the custom repository to use.
     type: str
+  username:
+    description:
+      - Username to authenticate against private repository.
```
jborean93, line 65: Should add `version_added: '1.4.0'` and `type: str` to these new options.

plugins/modules/win_psmodule.py
```diff
@@ -59,6 +60,12 @@
...
+  password:
+    description:
+      - Password to authenticate against private repository.
```
jborean93, line 68: Should add `version_added: '1.4.0'` and `type: str` to these new options.

plugins/modules/win_psrepository.py
```diff
@@ -38,6 +39,14 @@
     description:
       - Specifies the URI for publishing scripts to this repository.
     type: str
+  username:
+    description:
+      - Username to authenticate against private repository.
+    type: str
```
jborean93, line 45: Should add `version_added: '1.4.0'` to these new options.

plugins/modules/win_psrepository.py
```diff
@@ -38,6 +39,14 @@
...
+  password:
+    description:
+      - Password to authenticate against private repository.
+    type: str
```
jborean93, line 49: Should add `version_added: '1.4.0'` to these new options.

jborean93 (changes requested):
Apart from some minor docs stuff the changes seem ok here. It does look like the tests are now failing for both modules so that needs to be investigated before we can merge this in. My guess as to why it's failing is due to https://github.com/ansible-collections/community.windows/pull/199/files#diff-a35d8609fa2ed98c3244dd81bc151aaf708a3b6a6e0b2d030be94da2195e6989R159. It seems like `-Credential` is only valid for one parameter set and the other (when registering the default PSGallery one) does not have `-Credential`. It looks like you need to optionally set `-Credential` here. Could be as simple as adding it to `$changed_properties` based on whether SourceLocation, ScriptSourceLocation, PublishLocation, or ScriptPublishLocation is set.

It would also be good to have a changelog fragment for this change so people can see what has changed across the collection versions.

jborean93:
Closing due to the age and no reply from the review. Feel free to pick it up and reopen once they have been incorporated

Issue #211, opened 2021-03-09
win_pssession_configuration broken on devel


jborean93:
##### SUMMARY
It seems like https://github.com/ansible/ansible/pull/73760 has changed the async cleanup behaviour breaking this module on devel/2.11 (or at least on the tests.

Here is a test run with the failure https://dev.azure.com/ansible/community.windows/_build/results?buildId=8565&view=logs&j=f9fd81a9-8d8c-5970-ad60-a03da9d690c4&t=4cae795a-c3c8-5200-6b27-a9562cd144ad.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
win_pssession_configuration

##### ANSIBLE VERSION
```paste below
devel
```

PR #212, opened 2021-03-10
win_pssession_configuration - fix async_poll implementation and test

briantist:
##### SUMMARY

Fixes: #211

So this is interesting. A change in behavior in core around cleaning up after async tasks (https://github.com/ansible/ansible/pull/73760) seemed to be breaking the custom action in this module.

What it actually did was expose a bug in the implementation: when you set the `async_poll` option the action was never actually setting it on the task execution, so it always used the default polling of `15`. 

So the whole thing ends up running in polling mode anyway, and the action's loop of status checking only ever got invoked on the (inevitable) connection error that can't be handled properly by using `async_status` alone.

That fix is quite easy; what it then breaks is the "fire and forget" test, which.. should always have been broken (because if it wasn't, there'd be little need for the action other than convenience and check mode, since callers could use `async_status`).

~So I've updated the test to 1) actually ensure that the resulting `register:` result shows a status of `finished: 0` (this would have prevented this bug from being released in the first place), and 2) have enough stuff in the play tasks to actually wait for the result (hopefully). That last one may prove to be unstable in the future... I doubt it, but it could happen. If so, I have some remediation possibilities in the comments; in the worst case / quick fix that whole test case could be commented out or removed.~ 
this stuff is not all accurate, some of it is but I don't wanna edit it, just see later comments in this issue :-p

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

win_psession_configuration.py

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/212-win_pssession_configuration-async_poll.yml
+++ b/changelogs/fragments/212-win_pssession_configuration-async_poll.yml
@@ -1,3 +1,3 @@
 ---
 bugfixes:
-  - win_pssession_configuration - the ``async_poll`` option was not actually used and polling mode was always used with the default poll delay (https://github.com/ansible-collections/community.windows/pull/212).
+  - win_pssession_configuration - the ``async_poll`` option was not actually used and polling mode was always used with the default poll delay; this change also formally disables ``async_poll=0`` (https://github.com/ansible-collections/community.windows/pull/212).
--- a/changelogs/fragments/212-win_pssession_configuration-async_poll.yml
+++ b/changelogs/fragments/212-win_pssession_configuration-async_poll.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - win_pssession_configuration - the ``async_poll`` option was not actually used and polling mode was always used with the default poll delay (https://github.com/ansible-collections/community.windows/pull/212).
--- a/plugins/action/win_pssession_configuration.py
+++ b/plugins/action/win_pssession_configuration.py
@@ -56,6 +56,7 @@ class ActionModule(ActionBase):
         # if it's not in check mode, call the module async so the WinRM restart doesn't kill ansible
         if not check_mode:
             self._task.async_val = async_timeout
+            self._task.poll = async_poll
 
         result = status = self._execute_module(
             task_vars=task_vars,
--- a/plugins/action/win_pssession_configuration.py
+++ b/plugins/action/win_pssession_configuration.py
@@ -4,12 +4,10 @@
 from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
-import ntpath
 import time
 
 from ansible.errors import AnsibleError
 from ansible.plugins.action import ActionBase
-from ansible.utils.vars import merge_hash
 from ansible.playbook.task import Task
 from ansible.utils.display import Display
 display = Display()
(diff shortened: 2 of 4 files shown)


briantist:
womp, there's still problems with the test, definitely not stable, will see if there's anything I can do other than removing the test

briantist:
The "breakage" of properly using the value of `async_poll` and having it set to `0` is far more strange than it seemed. I'll try to document here. Long story short though: this PR will ensure the `async_poll` value is used, and will not allow it to be `0`.

---

For reasons beyond explanation or sense, when the action sets the poll value to `0` and then returns after `self._execute_module`, the _remote task_ fails, with an error somewhere deep in the `*PSSession*` command's code. Here's an example of the result file from a remote test server:
```json
{
    "ansible_async_watchdog_pid": 1800,
    "exception": "Cannot find path \u0027localhost\\Plugin\u0027 because it does not exist.\nAt line:98 char:3\r\n+   Get-ChildItem \u0027WSMan:\\localhost\\Plugin\\\u0027 -Force:$force | ? { $_.nam ...\r\n+   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~\n    + CategoryInfo          : ObjectNotFound: (localhost\\Plugin:String) [Get-ChildItem], ItemNotFoundException\n    + FullyQualifiedErrorId : PathNotFound,Microsoft.PowerShell.Commands.GetChildItemCommand\r\n\r\nScriptStackTrace:\r\nat \u003cScriptBlock\u003e, \u003cNo file\u003e: line 98\r\nat \u003cScriptBlock\u003e, \u003cNo file\u003e: line 95\r\nat Register-PSSessionConfiguration\u003cBegin\u003e, \u003cNo file\u003e: line 30\r\nat \u003cScriptBlock\u003e, \u003cNo file\u003e: line 217\r\nat \u003cScriptBlock\u003e, \u003cNo file\u003e: line 526\r\n\r\nSystem.Management.Automation.ItemNotFoundException: Cannot find path \u0027localhost\\Plugin\u0027 because it does not exist.\r\n   at System.Management.Automation.SessionStateInternal.GetChildItems(String path, Boolean recurse, UInt32 depth, CmdletProviderContext context)\r\n   at Microsoft.PowerShell.Commands.GetChildItemCommand.ProcessRecord()",
    "results_file": "C:\\Users\\Administrator\\.ansible_async\\849289196946.4880",
    "ansible_job_id": "849289196946.4880",
    "msg": "Unhandled exception while executing module: Cannot find path \u0027localhost\\Plugin\u0027 because it does not exist.",
    "started": 1,
    "_ansible_suppress_tmpdir_delete": true,
    "failed": true,
    "finished": 1
}
```

The strange part is, if I set the poll value to `0` intentionally, but _don't_ return from it from the action early, and let the action handle polling (instead of the task executor), the remote task does not fail. Similar to setting poll > 0, which lets the task executor poll (which eventually fails and then the action handles re-polling).

I know this doesn't make any sense, but I haven't been able to figure out why.

Anyway, it's moot with `async_poll=0` being unsupported; in reality that workflow wouldn't have been very useful anyway.

jborean93:
Thanks for fixing this @briantist!

PR #213, opened 2021-03-10
win_dhcp_lease: MAC address update to Docs

dmccuk:
##### SUMMARY

Using this module I was unable to create a DHCP reservation with a standard MAC address : seperator. Once I checked the code I could see it was looking for - and not : 

You can see it here:

https://github.com/ansible-collections/community.windows/blob/3afe02165344bf46534a9b56766e68531addac43/plugins/modules/win_dhcp_lease.ps1

line 124:

````
# MacAddress was specified
if ($mac) {
    if ($mac -like "*-*") {
        $mac_original = $mac
        $mac = Convert-MacAddress -mac $mac
    }
````

Once I made the update, the DHCP reservation worked.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

Document updates

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/docs/community.windows.win_dhcp_lease_module.rst
+++ b/docs/community.windows.win_dhcp_lease_module.rst
@@ -243,7 +243,7 @@ Examples
 
     - name: Ensure DHCP lease or reservation does not exist
       community.windows.win_dhcp_lease:
-        mac: 00:B1:8A:D1:5A:1F
+        mac: 00-B1-8A-D1-5A-1F
         state: absent
 
     - name: Ensure DHCP lease or reservation does not exist


jborean93:
The rst file you edited is dyanmically generated from the actual module docs at https://github.com/ansible-collections/community.windows/blob/main/plugins/modules/win_dhcp_lease.py. To edit them you should edit those files and the rst will be regenerated at the next release.

As for the problem at hand the link you specified is correct in that it's trying to match `*-*` but the code looks like it was meant to handle the other scenarios, it just doesn't do so properly. So while changing the docs to match what is currently working is great I hope at some future point in time that the module is updated to support both formats.

dmccuk:
Thanks. I'll see if I can update the actual code to cater for ":"

I'll close this request.

Issue #218, opened 2021-04-01
win_lineinfile trims trailing empty line


phoenikz:
Running a file through win_lineinfile and having it changed will result in a trailing empty row to be removed

This is because

    $before = [System.IO.File]::ReadAllLines($cleanpath, $encodingobj);

will not result in an empty last entry when the file at $cleanpath ends with \`n or \`r
and thus the later joining with $linesep will not keep the last line

PR #219, opened 2021-04-01
Enable keeping last empty line for win_lineinfile

phoenikz:
##### SUMMARY
Ensure that  win_lineinfile keeps empty last line
in the output file when changed.

Fixes: #218

##### ISSUE TYPE
- Bug fix

##### COMPONENT NAME
win_lineinfile

### DIFF
--- a/changelogs/fragments/win_lineinfile-newline.yml
+++ b/changelogs/fragments/win_lineinfile-newline.yml
@@ -1,0 +1,2 @@
+bugfixes:
+- win_lineinfile - Avoid stripping the newline at the end of a file - https://github.com/ansible-collections/community.windows/pull/219
--- a/plugins/modules/win_lineinfile.ps1
+++ b/plugins/modules/win_lineinfile.ps1
@@ -67,6 +67,7 @@ function Present($path, $regex, $line, $insertafter, $insertbefore, $create, $ba
 	# Note that we have to clean up the path because ansible wants to treat / and \ as
 	# interchangeable in windows pathnames, but .NET framework internals do not support that.
 	$cleanpath = $path.Replace("/", "\");
+    $endswithnewline = $null
 
 	# Check if path exists. If it does not exist, either create it if create == "yes"
 	# was specified or fail with a reasonable error message.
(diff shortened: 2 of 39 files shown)


jborean93:
Honestly I would probably prefer this be the behaviour and not exposed as an option. If we are editing a file then stripping the last line sounds like the wrong thing to do. It might be a breaking change but if someone is relying on this behaviour their setup is very brittle and they should ensure their source file contains what is needed as this is designed to edit a line.

We should also have some tests added for this scenario to ensure it doesn't regress in the future. We have a whole bunch of tests here https://github.com/ansible-collections/community.windows/blob/main/tests/integration/targets/win_lineinfile/tasks/main.yml. It would be great if you could add this scenario, maybe have a input file with 2 empty newlines at the end and make sure after `win_lineinfile` set what it needed to those 2 lines are still present.

phoenikz:
I have removed the option. This should break texts that rely on  https://github.com/ansible-collections/community.windows/blob/main/tests/integration/targets/win_lineinfile/files/test.txt, as the last row will be kept (it would not be previously). I will update checksums accordingly and possibly add another test too.

plugins/modules/win_lineinfile.py
```diff
@@ -94,6 +94,12 @@
     type: str
     choices: [ unix, windows ]
     default: windows
+  keepemptyendrow:
```
phoenikz, line 97: This will be removed in next commit, forgot

phoenikz:
Ready for re-review.
I had to change the implementation slightly since the creation of a new file (with create=yes) introduces one newline, and my
first solution saw doubling of newlines in some cases.

I also introduced expectation files to the tests to ease the translation of expectation to checksums to verify. The intention was to read checksum from the files, but they get copied with LF line-endings even though .gitattributes is set. One file even requires mixed line endings which is not well received by git unless the file is made binary.

I also added a test case for relative paths to win_lineinfile, as I have seen that the [System.IO.File] methods not always work as expected with relative paths.

jborean93 (approved):

jborean93:
Thanks for working on this and revamping the test cases. I made a minor change to keep using tabs which the module was already using. I also added a changelog fragment so the changes are present in the next changelog for others to see.

Issue #135, opened 2020-08-27
A module to change the drive letter and label


Udayendu:
##### SUMMARY
Currently we dont have a module that can be used to change the drive letter and label. `win_format` is there but it cant handle the label change as without formatting the drive.

##### ISSUE TYPE
- Feature Idea

##### ADDITIONAL INFORMATION
Currently I am using the following **win_shell** commands to change the drive letter and label:

- For Drive Letter Change:
```
    - name: Change the CD-ROM from E to H
      win_shell: |
        $cd = $NULL
        $cd = Get-WMIObject -Class Win32_CDROMDrive -ComputerName $env:COMPUTERNAME -ErrorAction Stop
        if ($cd.Drive -eq "E:")
        {
           Write-Output "Changing CD Drive letter from E: to H:"
           Set-WmiInstance -InputObject ( Get-WmiObject -Class Win32_volume -Filter "DriveLetter = 'E:'" ) -Arguments @{DriveLetter='h:'}
        }
```

- For Drive Label Change:
```
- name: Create new label for C drive
  win_shell: |
      Set-Volume -DriveLetter C -NewFileSystemLabel "OS"
```

jborean93:
The biggest issue with this request is changing something is not an idempotent operation. It's why Ansible has a copy but not a move module for files.

Udayendu:
> The biggest issue with this request is changing something is not an idempotent operation. It's why Ansible has a copy but not a move module for files.

Then whats the alternative. 

- Usually it should take a source drive letter that has to change and destination drive letter. If the destination is not used, it should change, else it should give a warning.
- For disk label I cant see any issue

jborean93:
I'm just merely stating that changing a letter is not an Ansible-like operation to do which is why it's not implemented. Typically we've just stated that if you want to do a destructive change (like a move) you implement it yourself as the conditions that people desire differ greatly so it's hard to please everyone.

I can see setting a disk label based on some other unique identifier being done but changing a disk label based on an existing disk label like the below is something that typically is reserved for manual win_shell commands

```yaml
- win_partition:
    original_label: D
    label: E
```

Udayendu:
> I'm just merely stating that changing a letter is not an Ansible-like operation to do which is why it's not implemented. Typically we've just stated that if you want to do a destructive change (like a move) you implement it yourself as the conditions that people desire differ greatly so it's hard to please everyone.
> 
[...]

PR #229, opened 2021-04-15
win_partition, win_format - Adding directory mountpoint handling

aelfwine88:
##### SUMMARY
New features added to the win_partition module:
- Attach partition to directory mountpoint
- Identify partition by directory mountpoint
- If not already exist create a directory for the directory mountpoint
- Remount partition
- Unmount partition
- Better error handling for drive_letter variable

Bugfix added to the win_partition module:
- Properly unmount partition before partition deletion

New feature added to the win_format module:
- Added a feature to identify volumes based on directory mountpoints

Changes in the python description files:
- Modified the description for the added features

##### ISSUE TYPE
- Feature Pull Request
- Bugfix Pull Request
- Docs Pull Request

##### COMPONENT NAME
- win_partition
- win_format

##### ADDITIONAL INFORMATION
I had to create a solution to attach partitions to directories in the filesystem so I modified @chopraaa's relevant modules.

With these changes the win_partition module is now able to identify or attach partitions by the file path of these directory mountpoints (eg.: E:\SQL_Log). This function requires an empty or a non existing directory. In later case the module will try to create the directory.

The module now also able to remount the selected partition to a different location or unmount it. Partition unmount can be triggered with setting the _drive_letter_ to **none**.

In every scenario when the module caries out a mounting all the previous mountpoints are getting wiped out. This means with this module it is only possible to mount a partition to one location (drive letter or mountpoint).

I also added some additional error handing / formating to the _drive_letter_ variable to be able to accept root paths like C:\ instead of only letters like C.

Additional bugfix was introduced in this module as well: in case of a partition deletion now the module will try to unmount the partition prior the deletion. This prevents bugged/partially locked drive letter or directory mountpoint residues in the OS.

The win_format module was also upgraded to be able to identify volumes based on directory mountpoints. This ensures that you can format volumes based on this attribute as well (eg.: formating partition attached under E:\SQL_Log).

Naturally I completed the relevant two python man files (win_partition.py, win_format.py) with these changes.

### DIFF
--- a/plugins/modules/win_format.ps1
+++ b/plugins/modules/win_format.ps1
@@ -19,21 +19,29 @@ $spec = @{
         file_system = @{ type = "str"; choices = "ntfs", "refs", "exfat", "fat32", "fat" }
         allocation_unit_size = @{ type = "int" }
         large_frs = @{ type = "bool" }
-        full = @{ type = "bool"; default = $false }
-        compress = @{ type = "bool" }
+        full_format = @{ type = "bool"; default = $false }
+        compress_volume = @{ type = "bool" }
         integrity_streams = @{ type = "bool" }
         force = @{ type = "bool"; default = $false }
+        filepath = @{ type = "str" }
     }
     mutually_exclusive = @(
-        ,@('drive_letter', 'path', 'label')
+        ,@('drive_letter', 'path', 'label', 'filepath')
     )
     required_one_of = @(
-        ,@('drive_letter', 'path', 'label')
+        ,@('drive_letter', 'path', 'label', 'filepath')
     )
     supports_check_mode = $true
 }
 
 $module = [Ansible.Basic.AnsibleModule]::Create($args, $spec)
+
+$filepath = $null
+
+if(Test-Path $drive_letter -PathType Container) {
+    $filepath = $drive_letter
+    $drive_letter = $null
+}
 
 $drive_letter = $module.Params.drive_letter
 $path = $module.Params.path
@@ -69,7 +77,8 @@ function Get-AnsibleVolume {
     param(
         $DriveLetter,
         $Path,
-        $Label
+        $Label,
+        $FilePath
     )
 
     if ($null -ne $DriveLetter) {
@@ -93,8 +102,15 @@ function Get-AnsibleVolume {
             $module.FailJson("There was an error retrieving the volume using label $($Label): $($_.Exception.Message)", $_)
         }
     }
+    elseif ($null -ne $FilePath) {
+        try {
+            $volume = Get-Volume -FilePath $FilePath
+        } catch {
+            $module.FailJson("There was an error retrieving the volume using filepath $($FilePath): $($_.Exception.Message)", $_)
+        }
+    }
     else {
-        $module.FailJson("Unable to locate volume: drive_letter, path and label were not specified")
+        $module.FailJson("Unable to locate volume: drive_letter, path, label or filepath were not specified")
     }
 
     return $volume
@@ -138,7 +154,7 @@ function Format-AnsibleVolume {
 
 }
 
-$ansible_volume = Get-AnsibleVolume -DriveLetter $drive_letter -Path $path -Label $label
+$ansible_volume = Get-AnsibleVolume -DriveLetter $drive_letter -Path $path -Label $label -FilePath $filepath
 $ansible_file_system = $ansible_volume.FileSystem
 $ansible_volume_size = $ansible_volume.Size
 $ansible_volume_alu = (Get-CimInstance -ClassName Win32_Volume -Filter "DeviceId = '$($ansible_volume.path.replace('\','\\'))'" -Property BlockSize).BlockSize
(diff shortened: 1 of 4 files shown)


aelfwine88:
Also solves this issue with an idempotent way:
https://github.com/ansible-collections/community.windows/issues/135

plugins/modules/win_partition.py
```diff
@@ -91,9 +97,9 @@
     partition_size: 5 GiB
     disk_number: 1
 
-- name: Resize previously created partition to it's maximum size and change it's drive letter to E
+- name: Resize previously created partition to it's maximum size and change it's mountpoint to E:\SQL_Log
   community.windows.win_partition:
-    drive_letter: E
+    drive_letter: E:\SQL_Log
```
rgl, line 102: since this is now a mountpoint, should `drive_letter` be renamed? or a new `mountpoint` variable added?

aelfwine88: I think replacing the `drive_letter` variable with `mountpoint` is the better choice since drive letters are actually mountpoints as well.

aelfwine88: Should I upload a modified version based on this?

rgl: Removing the `driver_letter` seems to be a breaking change, so at least, the module major version needs to be bumped if `driver_letter` is removed. Maybe supporting the two is more appropriate. And maybe there's a way to automatically alias `drive_letter` to `mountpoint`?

Please note that I'm not a maintainer on this repository. I just started using this today :-)

aelfwine88: Then the maintainers will decide I guess. Frankly, I don't really see the point of adding an alias that logically does the exact same thing, but I see your point that with this feature added the `drive_letter` variable name is under-representing the option it is used for. At the other hand a replace could be easily done with a ctrl+h and it wouldn't need any additional change in the code or in the documentation.

plugins/modules/win_format.ps1
```diff
@@ -19,22 +19,30 @@ $spec = @{
         file_system = @{ type = "str"; choices = "ntfs", "refs", "exfat", "fat32", "fat" }
         allocation_unit_size = @{ type = "int" }
         large_frs = @{ type = "bool" }
-        full = @{ type = "bool"; default = $false }
-        compress = @{ type = "bool" }
+        full_format = @{ type = "bool"; default = $false }
```
jborean93, line 22: We can't change the existing options, if you wish to create ones that are a bit more explanatory then you should add an `alias` instead.

```suggestion
        full = @{ type = "bool"; default = $false; aliases = @(,'full_format') }
```

plugins/modules/win_format.ps1
```diff
@@ -19,22 +19,30 @@ $spec = @{
...
+        ,@('drive_letter', 'path', 'label', 'filepath')
     )
     supports_check_mode = $true
 }
 
 $module = [Ansible.Basic.AnsibleModule]::Create($args, $spec)
 
+$filepath = $null
+
+if(Test-Path $drive_letter -PathType Container) {
```
jborean93, line 41: `$drive_letter` has been defined yet so this will never be `$true`. Looks like you want to put this further below.

plugins/modules/win_format.ps1
```diff
@@ -19,22 +19,30 @@ $spec = @{
...
     required_one_of = @(
-        ,@('drive_letter', 'path', 'label')
+        ,@('drive_letter', 'path', 'label', 'filepath')
     )
     supports_check_mode = $true
 }
 
 $module = [Ansible.Basic.AnsibleModule]::Create($args, $spec)
 
+$filepath = $null
```
jborean93, line 39: You aren't setting `$filepath` to anything, you want to add this below with the other options with `$filepath = $module.Params.filepath`.

plugins/modules/win_format.py
```diff
@@ -12,7 +12,7 @@
 options:
   drive_letter:
     description:
-      - Used to specify the drive letter of the volume to be formatted.
+      - Used to specify the drive letter or directory mountpoint of the volume to be formatted.
```
jborean93, line 15: You haven't documented the new module option you've added (`filepath`). It sounds like the directory mountpoint is specified there instead.

plugins/modules/win_partition.ps1
```diff
@@ -1,6 +1,7 @@
 #!powershell
 
 # Copyright: (c) 2018, Varun Chopra (@chopraaa) <v@chopraaa.com>
+# Mountpoint feature added by Eriol (@aelfwine88)
```
jborean93, line 4: This isn't needed, we use the git attribution history to determine this information.

jborean93:
I've just done a very basic review on the `win_format` side. Please make sure you test this logic, it would be even better if you could add the tests to our test suite so they run in CI. There's a few things in there that don't seem like it would work.

There's also a whole bunch of failures in CI that need to be addressed before we can continue forward.

plugins/modules/win_partition.py
jborean93: We shouldn't make a breaking change but what we can do is change the option to `mountpoint` to better reflect what it is used for but add an alias for `drive_letter` so that existing playbooks will continue to work. The alias is key to ensuring that people using this module don't need to deal with a breaking change, especially one where there's no in between where both options work.

jborean93:
Closing due to the age of the PR and no responses since the last review. If you, or someone else, wishes to try and continue forward please open a new PR with the changes requested.

PR #231, opened 2021-04-16
Hide password for win_psexec module

### LINKED ISSUE #43: win_psexec module is exposing password to output
_From @cuichenli on Nov 18, 2019 05:04_

##### SUMMARY
The `win_psexec` module is exposing the password and username to output, is that intended? Is there a way to hide it? Thanks!

My playbook:
```
- name: Psexec
  win_psexec:
    command: >
      powershell.exe 
    session: 1
    wait: yes
    interactive: yes
    username: "username"
    password: "password"
```
 
```
changed: [MYSERVER] => {
	....
    "psexec_command": "psexec64.exe -u username -p password -i 1 -accepteula powershell.exe",
  	....
```
##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
win_psexec

##### ANSIBLE VERSION

```paste below
ansible 2.8.5
python version = 3.7.3 (default, Apr  3 2019, 05:39:12) [GCC 8.3.0]
```

##### OS / ENVIRONMENT

Host OS : Linux
Target OS: Windows 

_Copied from original issue: ansible/ansible#64978_

quadzz:
##### SUMMARY
According to #43 - we do not want to show password when executing win_psexec. This change allows user to explicitly show it (for example when testing) while hiding the password by default

Fixes #43 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_psexec

##### ADDITIONAL INFORMATION
Added new parameter log_password

### DIFF
--- a/changelogs/fragments/win_psexec-cmd-output.yml
+++ b/changelogs/fragments/win_psexec-cmd-output.yml
@@ -1,0 +1,2 @@
+security_fixes:
+- win_psexec - Ensure password is masked in ``psexec_command`` return result - https://github.com/ansible-collections/community.windows/issues/43
--- a/plugins/modules/win_psexec.ps1
+++ b/plugins/modules/win_psexec.ps1
@@ -27,6 +27,7 @@ $spec = @{
         session = @{ type='int' }
         priority = @{ type='str'; choices=@( 'background', 'low', 'belownormal', 'abovenormal', 'high', 'realtime' ) }
         timeout = @{ type='int' }
+        log_password = @{ type='bool'; default=$false }
     }
 }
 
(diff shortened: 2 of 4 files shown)


jborean93:
Personally I think we should just mask the password all the time, there's very little reason why people should see the password returned back in the argument command.

quadzz:
@jborean93 yes I definitely agree. 
That's why I included an additional parameter log_password which actually shows password, since someone may want to show the password (e.g. when debugging)
Otherwise it's always replaced with "*PASSWORD_REPLACED*" string.

jborean93:
I'm saying that the password should be escaped all the time, there's no need for it to show up in the logs. The only other issue is the code needs to be adjusted so that it can replace passwords that contain `"` in them as the argument conversion will transform `test"ing` to `test\"ing`.

jborean93:
I'm hoping to get this into the next `community.windows` release so I've added some changes on top of yours to fit my last comment. Thanks for your work on this.

Issue #235, opened 2021-04-28
need win_zip or win_archive module 


sugitk:
##### SUMMARY

We cannot create a zip archive on Windows

##### ISSUE TYPE
- Feature Idea

##### ADDITIONAL INFORMATION

```yaml

```

mariairisong:
Hi, looking for win_zip for ansible 2.9, similar to the module in ansible 2.8. Thanks! While win_shell works, we're looking for this module so playbooks are the same across different ansible servers.

hiyokotaisa:
New module has been merged so I think we can close this issue

PR #236, opened 2021-04-30
New module: win_zip

hiyokotaisa:
##### SUMMARY

Add win_zip module to fix issue #235 

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

- win_zip

### DIFF
--- a/changelogs/fragments/235-add-win_zip.yml
+++ b/changelogs/fragments/235-add-win_zip.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+- win_zip - new module
--- a/changelogs/fragments/235-add-win_zip.yml
+++ b/changelogs/fragments/235-add-win_zip.yml
@@ -1,3 +1,0 @@
----
-minor_changes:
-- win_zip - new module
--- a/plugins/modules/win_zip.ps1
+++ b/plugins/modules/win_zip.ps1
@@ -1,0 +1,57 @@
+#!powershell
+
+# Copyright: (c) 2021, Kento Yagisawa <thel.vadam2485@gmail.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+#Requires -Module Ansible.ModuleUtils.Legacy
+
+$ErrorActionPreference = "Stop"
+
+$params = Parse-Args $args -supports_check_mode $true
+$check_mode = Get-AnsibleParam -obj $params -name "_ansible_check_mode" -type "bool" -default $false
+
+$src = Get-AnsibleParam -obj $params -name "src" -type "path" -failifempty $true
+$dest = Get-AnsibleParam -obj $params -name "dest" -type "path" -failifempty $true
+$force = Get-AnsibleParam -obj $params -name "overwrite" -type "bool" -default $false
+
+$result = @{
+    changed = $false
+    dest = $dest -replace '\$',''
+    src = $src -replace '\$',''
+}
+
+If(-not (Test-Path -LiteralPath $src)) {
+    Fail-Json -obj $result -message "The source file or directory '$src' does not exist."
+}
+
+If($dest -notlike "*.zip") {
+    Fail-Json -obj $result -message "The destination zip file path '$dest' need to be zip file path."
+}
+
+If((Test-Path -LiteralPath $dest) -and ($force -eq $false)) {
+    $result.skipped = $true
+    $result.msg = "The destination zip file '$dest' already exists."
+    Exit-Json -obj $result
+}
+
+Function Compress-Zip($src, $dest) {
+    try {
+        If((Test-Path -LiteralPath $dest) -and ($force -eq $true) -and (-not $check_mode)) {
+            Remove-Item -LiteralPath $dest
+        }
+        If (-not $check_mode) {
+            Compress-Archive -LiteralPath $src -DestinationPath $dest
+        }
+        $result.changed = $true
+    } catch {
+        Exit-Json -obj $result
+    }
+}
+
+try {
+    Compress-Zip -src $src -dest $dest
+    } catch {
+        Fail-Json -obj $result -message "There is an unexpected error during the zip compression process."
+    }
+
+Exit-Json $result
--- a/plugins/modules/win_zip.ps1
+++ b/plugins/modules/win_zip.ps1
@@ -3,55 +3,57 @@
 # Copyright: (c) 2021, Kento Yagisawa <thel.vadam2485@gmail.com>
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
-#Requires -Module Ansible.ModuleUtils.Legacy
+#AnsibleRequires -CSharpUtil Ansible.Basic
 
-$ErrorActionPreference = "Stop"
-
-$params = Parse-Args $args -supports_check_mode $true
-$check_mode = Get-AnsibleParam -obj $params -name "_ansible_check_mode" -type "bool" -default $false
-
-$src = Get-AnsibleParam -obj $params -name "src" -type "path" -failifempty $true
-$dest = Get-AnsibleParam -obj $params -name "dest" -type "path" -failifempty $true
-$force = Get-AnsibleParam -obj $params -name "overwrite" -type "bool" -default $false
-
-$result = @{
-    changed = $false
-    dest = $dest -replace '\$',''
-    src = $src -replace '\$',''
+$spec = @{
+    options = @{
+        src = @{ type = 'path'; required = $true }
+        dest = @{ type = 'path'; required = $true }
+    }
+    supports_check_mode = $true
 }
+$module = [Ansible.Basic.AnsibleModule]::Create($args, $spec)
+
+$src = $module.Params.src
+$dest = $module.Params.dest
+
+$srcFile = [System.IO.Path]::GetFileName($src)
+$compressionLevel = [System.IO.Compression.CompressionLevel]::Optimal
 
 If(-not (Test-Path -LiteralPath $src)) {
-    Fail-Json -obj $result -message "The source file or directory '$src' does not exist."
+    $module.FailJson("The source file or directory '$src' does not exist.")
 }
 
 If($dest -notlike "*.zip") {
-    Fail-Json -obj $result -message "The destination zip file path '$dest' need to be zip file path."
+    $module.FailJson("The destination zip file path '$dest' need to be zip file path.")
 }
 
-If((Test-Path -LiteralPath $dest) -and ($force -eq $false)) {
-    $result.skipped = $true
-    $result.msg = "The destination zip file '$dest' already exists."
-    Exit-Json -obj $result
+If(Test-Path -LiteralPath $dest) {
+    $module.Result.skipped = $true
+    $module.Result.msg = "The destination zip file '$dest' already exists."
+    $module.ExitJson()
+}
+
+# Check .NET v4.5 or later version exists or not
+If(-not (Test-Path -LiteralPath 'HKLM:\SOFTWARE\Microsoft\NET Framework Setup\NDP\v4\Full')) {
+    $module.FailJson(".NET Framework 4.5 or later version needs to be installed.")
 }
 
 Function Compress-Zip($src, $dest) {
-    try {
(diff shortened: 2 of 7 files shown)


changelogs/fragments/235-add-win_zip.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
```
jborean93, line 2: Changelog fragments for new modules aren't needed. It is automatically added when the changelog is generated.

plugins/modules/win_zip.ps1
```diff
@@ -0,0 +1,57 @@
...
+
+$src = Get-AnsibleParam -obj $params -name "src" -type "path" -failifempty $true
+$dest = Get-AnsibleParam -obj $params -name "dest" -type "path" -failifempty $true
+$force = Get-AnsibleParam -obj $params -name "overwrite" -type "bool" -default $false
+
+$result = @{
+    changed = $false
+    dest = $dest -replace '\$',''
+    src = $src -replace '\$',''
+}
```
jborean93, line 21: We should try and use the new module wrapper for this instead of the legacy one. There are numerous benefits you get from using that over the older style

```suggestion
#AnsibleRequires -CSharpUtil Ansible.Basic

$spec = @{
    options = @{
        src = @{ type = 'path'; required = $true }
        dest = @{ type = 'path'; required = $true }
        overwrite = @{ type = 'bool'; default = $false }
    }
    supports_check_mode = $true
}
$module = [Ansible.Basic.AnsibleModule]::Create($args, $spec)

$src = $module.Params.src
$dest = $module.Params.dest
$force = $module.Params.overwrite
```

Some of the features are

* Automatic module invocation details when run with `-vvv`
    * You no longer need to, or should, return `src` and `dest` in your return value
* Better option validation and error messages around that
* Event logging when running the modules

Some things you do need to change with the newer format

* `Fail-Json` becomes `$module.FailJson("My message here")`
* `Exit-Json` become `$module.ExitJson()`
* Getting check mode is now `$module.CheckMode
* Settings results is now done on `$module.Results.foo = 'bar'`, e.g. `$module.Results.changed = $true`

plugins/modules/win_zip.ps1
```diff
@@ -0,0 +1,57 @@
...
+    try {
+        If((Test-Path -LiteralPath $dest) -and ($force -eq $true) -and (-not $check_mode)) {
+            Remove-Item -LiteralPath $dest
+        }
+        If (-not $check_mode) {
+            Compress-Archive -LiteralPath $src -DestinationPath $dest
+        }
+        $result.changed = $true
+    } catch {
+        Exit-Json -obj $result
```
jborean93, line 47: This will make any errors when you go to compress the zip. You may as well just remove the `try { } catch { }` altogether as you are catching it when calling `Compress-Zip`.

plugins/modules/win_zip.ps1
```diff
@@ -0,0 +1,57 @@
...
+    }
+}
+
+try {
+    Compress-Zip -src $src -dest $dest
+    } catch {
+        Fail-Json -obj $result -message "There is an unexpected error during the zip compression process."
+    }
+
+Exit-Json $result
```
jborean93, line 57: When using the new module wrapper this would now become

```suggestion
$module.ExitJson()
```

plugins/modules/win_zip.ps1
```diff
@@ -0,0 +1,57 @@
...
+        $result.changed = $true
+    } catch {
+        Exit-Json -obj $result
+    }
+}
+
+try {
+    Compress-Zip -src $src -dest $dest
+    } catch {
+        Fail-Json -obj $result -message "There is an unexpected error during the zip compression process."
```
jborean93, line 54: It's better if you were to just let the exception go uncaught. We will pick it up automatically and generate a nice error message with a stack trace for you. Unless there's a pressing need to catch specific errors I recommend you don't.

plugins/modules/win_zip.py
```diff
@@ -0,0 +1,69 @@
...
+    description:
+      - File or directory path to be zipped (provide absolute path on the target node).
+    type: path
+    required: yes
+  dest:
+    description:
+      - Destination path of zip file (provide absolute path of zip file on the target node).
+    type: path
+    required: yes
+  overwrite:
```
jborean93, line 27: Personally I don't think we should have this. If someone truly wants to always compress the src then they should have a win_file task beforehand to delete it like

```yaml
- win_file:
    path: C:\my\file.zip
    state: absent
```

This allows us to slowly add idempotency to the module that's a lot smarter than just checking if the dest file exists or not.

plugins/modules/win_zip.py
```diff
@@ -0,0 +1,69 @@
...
+    description:
+      - If the zip file exists, the task will delete the old zip file then recreate it.
+    type: bool
+    default: false
+notes:
+- The maximum file size is 2GB because there's a limitation of the underlying API.
```
jborean93, line 33: Technically if you use the pure .NET way you can bypass this limit. If you are interested in having that in the module let me know and I can share the details. The other benefit is you aren't limited to just PowerShell v5. The .NET way is possible if you have .NET Framework v4.5 installed.

plugins/modules/win_zip.py
```diff
@@ -0,0 +1,69 @@
...
+
+- name: Recreate a zip file when the dest path already exists
+  community.windows.win_zip:
+    src: C:\Users\hiyoko\log.txt
+    dest: C:\Users\hiyoko\log.zip
+    overwrite: yes
+'''
+
+RETURN = r'''
+dest:
```
jborean93, line 59: Personally I don't think you should return these, use the new module wrapper and it's done automatically and doesn't need to be documented.

jborean93 (changes requested):
Thank you very much for adding the tests, it is greatly appreciated. I've added a few comments on the module itself which I feel like we should implement before merging it in. If you have any questions feel free to let me know.

changelogs/fragments/235-add-win_zip.yml
hiyokotaisa: Thanks! I will remove it then

plugins/modules/win_zip.py
hiyokotaisa: Sure, I will remove it

plugins/modules/win_zip.py
hiyokotaisa: Thank you for the advice! I will rewrite those with .NET way.

plugins/modules/win_zip.py
hiyokotaisa: Yeah, I will remove `overwrite` option then.

plugins/modules/win_zip.ps1
hiyokotaisa: yeah that was unnecessary try catch sentence... thank you for the advice!

plugins/modules/win_zip.ps1
hiyokotaisa: Okay I will rewrite the module with the new module wrapper

plugins/modules/win_zip.ps1
hiyokotaisa: yeah that will work like black hole... I will remove it thanks!

plugins/modules/win_zip.ps1
```diff
@@ -0,0 +1,59 @@
...
+If(-not (Test-Path -LiteralPath 'HKLM:\SOFTWARE\Microsoft\NET Framework Setup\NDP\v4\Full')) {
+    $module.FailJson(".NET Framework 4.5 or later version needs to be installed.")
+}
+
+Function Compress-Zip($src, $dest) {
+    If (-not $module.CheckMode) {
+        Add-Type -AssemblyName System.IO.Compression.FileSystem
+
+        If (Test-Path -LiteralPath $src -PathType Container) {
+            $zip = [System.IO.Compression.ZipFile]::CreateFromDirectory($src, $dest, $compressionLevel, $true)
```
jborean93, line 47: It would be nice to set the last boolean depending is `$src` ends with `\*` or not. Basically if someone does `src: C:\dir\*` then you don't want to include the base directory (`$false`). If it is just `src: C:\dir` then you want to include the base directory (`$true`).

A minor nit is that `CreateFromDirectory` is a void method and doesn't return anything. You don't need the `$zip =`

plugins/modules/win_zip.ps1
```diff
@@ -0,0 +1,59 @@
...
+        } Else {
+            $zip = [System.IO.Compression.ZipFile]::Open($dest, 'Update')
+            [System.IO.Compression.ZipFileExtensions]::CreateEntryFromFile($zip, $src, $srcFile, $compressionLevel)
+            $zip.Dispose()
```
jborean93, line 51: You will want to wrap this in a try finally to ensure the zip is disposed even on an error.

```suggestion
            $zip = [System.IO.Compression.ZipFile]::Open($dest, 'Update')
            try {
                [System.IO.Compression.ZipFileExtensions]::CreateEntryFromFile($zip, $src, $srcFile, $compressionLevel)
            } finally {
                $zip.Dispose()
            }
```

tests/integration/targets/win_zip/tasks/main.yml
```diff
@@ -0,0 +1,113 @@
...
+    path: C:\testdir\src\
+    state: directory
+
+- name: create testdir\zipped directory for CI
+  ansible.windows.win_file:
+    path: C:\testdir\zipped\
+    state: directory
+
+- name: create test files for CI
+  ansible.builtin.win_shell: 'New-Item -Force -Path C:\testdir\src\test.txt -Value "This is a test file."'
```
jborean93, line 18: You want to use `Set-Content` here instead of `New-Item`.

hiyokotaisa:
Thank you for the review @jborean93 ! :smile: 
I modified the module according to your comment, CI had been failed because of `New-Item` cmdlet but it seems okay after I changed it to `Set-Content`.

plugins/modules/win_zip.ps1
```diff
@@ -49,11 +52,11 @@ try {
 Function Compress-Zip($src, $dest) {
     If (-not $module.CheckMode) {
         If (Test-Path -LiteralPath $src -PathType Container) {
-            [System.IO.Compression.ZipFile]::CreateFromDirectory($src, $dest, $compressionLevel, $srcWildcard)
+            [System.IO.Compression.ZipFile]::CreateFromDirectory($src, $dest, $compressionLevel, (-not $srcWildcard), $encoding)
         } Else {
             $zip = [System.IO.Compression.ZipFile]::Open($dest, 'Update')
             try {
-                [System.IO.Compression.ZipFileExtensions]::CreateEntryFromFile($zip, $src, $srcFile, $compressionLevel)
+                [void][System.IO.Compression.ZipFileExtensions]::CreateEntryFromFile($zip, $src, $srcFile, $compressionLevel)
```
jborean93, line 59: `CreateEntryFromFile` outputs data, to avoid it clashing with the module return json and showing an error we just void it out.

tests/integration/targets/win_zip/tasks/main.yml
```diff
@@ -1,33 +1,54 @@
 ---
-- name: clean up the old dir
-  ansible.windows.win_file:
-    path: C:\testdir\
-    state: absent
+- set_fact:
```
jborean93, line 2: This is to just expand the tests a bit more and validate it zipped the files as we expected.

tests/integration/targets/win_zip/meta/main.yml
```diff
@@ -0,0 +1,2 @@
+dependencies:
+- setup_remote_tmp_dir
```
jborean93, line 2: This handles the creation and deletion of temp dirs in tests, you can access them with `{{ remote_tmp_dir }}` as done in `defaults/main.yml`.

tests/integration/targets/win_zip/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+win_zip_name: .ÅÑŚÌβŁÈ [$!@^&test(;)] 👋
```
jborean93, line 1: Using some non-ASCII characters here to test out some edge cases when it comes to Windows. This ensures that the module is able to handle these non-ASCII chars and verifies the UTF-8 encoding we specify is actually in affect.

jborean93:
Ignoring Zuul check as it's a known failure and unrelated to any changes here.

PR #248, opened 2021-05-14
streamline AD user attribute setting

conitrade-as:
##### SUMMARY
This change set ensures that running in change mode does return the user
object as well. Otherwise, Ansible reports the user as being absent. Which is
pretty confusing for the caller.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_domain_user

##### ADDITIONAL INFORMATION
N/A


jborean93 (approved):
Thanks for the fix

Issue #247, opened 2021-05-14
win_product_facts: The module doesn't return correctly the product id


sky-joker:
##### SUMMARY

I got the product id of Windows 10 with the win_product_facts module after I installed a KMS client key to Windows 10, but the id was wrong.
I installed the key is `W269N-WFGWX-YVC9B-4J6C9-T83GX`, but I actually got `R3HB7-X6HMH-F48J3-8XJVP-9QBP9`.
https://docs.microsoft.com/en-us/windows-server/get-started/kmsclientkeys#windows-10-all-supported-semi-annual-channel-versions

That is the following result.

```
192.168.10.63 | SUCCESS => {
    "ansible_facts": {
        "ansible_os_license_channel": "Volume:GVLK",
        "ansible_os_license_edition": "Windows(R), Professional edition",
        "ansible_os_license_status": "Notification",
        "ansible_os_product_id": "00331-10000-00001-AA171",
        "ansible_os_product_key": "R3HB7-X6HMH-F48J3-8XJVP-9QBP9"
    },
    "changed": false
}
```

I seem that the cause of the issue is the converting process of the product id has an issue.
https://github.com/ansible-collections/community.windows/blob/ca03e76746dcd0923f1cffb9c2a893a6f89908fa/plugins/modules/win_product_facts.ps1#L39-L57

##### ISSUE TYPE

- Bug Report

##### COMPONENT NAME

plugins/modules/win_product_facts.ps1

##### ANSIBLE VERSION

```
# ansible --version
ansible 2.10.9
```

##### CONFIGURATION

```
# ansible-config dump --only-changed
COLLECTIONS_PATHS(/root/community.windows/PR/xxx/ansible.cfg) = ['/root/community.windows/PR/xxx/collections']
```

##### OS / ENVIRONMENT

Windows 10 Professional

##### STEPS TO REPRODUCE

I executed the following command.

```
# ansible all -i inventory.yml -m community.windows.win_product_facts -u administrator -k
SSH password:
```

##### EXPECTED RESULTS

The module gets correctly the product id.

##### ACTUAL RESULTS

The module doesn't get correctly the product id.

PR #251, opened 2021-05-15
win_product_facts: fix an issue that the module doesn't correctly convert a product id

sky-joker:
##### SUMMARY

The win_product_facts module has an issue that doesn't convert to the correct id from the registered product id in Windows10.(maybe a bug)
This PR will fix the issue.

fixes: https://github.com/ansible-collections/community.windows/issues/247

I made the patch while referencing the following vbs(productkey.vbs).

https://answers.microsoft.com/en-us/windows/forum/windows_other-windows_install/get-preinstalled-product-key/e39ec36f-ec3b-4495-b049-be02527a21ce  
https://answers.microsoft.com/en-us/windows/forum/windows_10-windows_install-winpc/windows-10-activation-error-0x8007007b/45d12b1c-30b1-4001-bc10-0bcd3da7db1f

By the way, the product id correctly gets with the productkey.vbs.

##### COMPONENT NAME

plugins/modules/win_product_facts.ps1

##### ADDITIONAL INFORMATION

Tested on Windows 10.

### DIFF
--- a/changelogs/fragments/251-win_product_facts.yml
+++ b/changelogs/fragments/251-win_product_facts.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - win_product_facts - fixed an issue that the module doesn't correctly convert a product id (https://github.com/ansible-collections/community.windows/pull/251).
--- a/plugins/modules/win_product_facts.ps1
+++ b/plugins/modules/win_product_facts.ps1
@@ -38,6 +38,9 @@ if (-not $product_key) {
 
     if ($data) {
         $product_key = $null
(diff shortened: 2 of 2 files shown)


sky-joker:
Unfortunately, I confirmed the module with the patch doesn’t correctly convert to a string in a part of MAK.
By the way, I also confirmed the existing module doesn’t also correctly convert even with the same MAK.

jborean93 (approved):

jborean93:
Thanks for fixing this, I've verified this with some eval ISOs and the last digits now match what Windows reports in it's settings.

Issue #259, opened 2021-06-02
win_dns_zone does not handle forwarders correctly


BoldlyGoingNoWhre:
##### SUMMARY
Plays that have forwarders (non replicated) in them are not idempotent and fail to update

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
win_dns_zone

##### ANSIBLE VERSION

```paste below
ansible 2.9.6
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/pvandepu/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.8.5 (default, Jan 27 2021, 15:41:15) [GCC 9.3.0]
```

##### CONFIGURATION

```paste below
n/a
```

##### OS / ENVIRONMENT

target OS Windows 2021 R2

##### STEPS TO REPRODUCE

```yaml
- name: Create DNS forwarder
      community.windows.win_dns_zone:
        name: myowndnszone.loc
        type: forwarder
        dns_servers:
          - 1.1.1.1
          - 2.2.2.2
```

##### EXPECTED RESULTS

DNS conditional forwarder zone is created / updated / unchanged if OK

##### ACTUAL RESULTS

DNS conditional forwarder zone is created if not exists (OK)
running the same play a second time results in error, running play with changed values results in error

```paste below
FAILED! => {"changed": false, "msg": "Failed to set properties on the zone myowndnszone.loc: Cannot validate argument on parameter 'ReplicationScope'. The argument \"None\" does not belong to the set \"Forest,Domain,Legacy,Custom\" specified by the ValidateSet attribute. Supply an argument that is in the set and then try the command again."}
```

jborean93:
This is fixed with https://github.com/ansible-collections/community.windows/pull/260.

PR #260, opened 2021-06-03
fix forwarder replicationscope none

BoldlyGoingNoWhre:
##### SUMMARY

Fixes issue #259

plays with forwarders without replication are now idempotent & updatable. 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

win_dns_zone

##### ADDITIONAL INFORMATION

```paste below
Dropping parameter ReplicationScope from processing if it was already set to none for conditional forwarding dns zones. 
```


jborean93 (approved):
Thanks for the fix, I've pushed a commit which adds a changelog fragment just so others can see what has changed in the next `community.windows` release.

Issue #256, opened 2021-05-29
win_domain_group_membership: Group members can't be set when state is pure


CathalMullan:
##### SUMMARY
When setting the members of a group using `win_domain_group_membership`, they are added then removed within the same task.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
win_domain_group_membership

##### ANSIBLE VERSION
```paste below
ansible [core 2.11.1]
  config file = /home/cathal/workspace/ansible/ansible.cfg
  configured module search path = ['/home/cathal/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/cathal/workspace/ansible/.venv/lib/python3.9/site-packages/ansible
  ansible collection location = /home/cathal/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/cathal/workspace/ansible/.venv/bin/ansible
  python version = 3.9.2 (default, Mar 24 2021, 13:57:04) [GCC 7.5.0]
  jinja version = 3.0.1
  libyaml = True
```

##### CONFIGURATION
```paste below
DEFAULT_FORKS(/home/cathal/workspace/ansible/ansible.cfg) = 50
DEFAULT_GATHERING(/home/cathal/workspace/ansible/ansible.cfg) = smart
DEFAULT_HOST_LIST(/home/cathal/workspace/ansible/ansible.cfg) = ['/home/cathal/workspace/ansible/vagrant.ini']
DEFAULT_LOAD_CALLBACK_PLUGINS(/home/cathal/workspace/ansible/ansible.cfg) = True
DEFAULT_POLL_INTERVAL(/home/cathal/workspace/ansible/ansible.cfg) = 3
DEFAULT_ROLES_PATH(/home/cathal/workspace/ansible/ansible.cfg) = ['/home/cathal/workspace/ansible/roles']
DEFAULT_STDOUT_CALLBACK(/home/cathal/workspace/ansible/ansible.cfg) = yaml
DEPRECATION_WARNINGS(/home/cathal/workspace/ansible/ansible.cfg) = True
DISPLAY_SKIPPED_HOSTS(/home/cathal/workspace/ansible/ansible.cfg) = False
HOST_KEY_CHECKING(/home/cathal/workspace/ansible/ansible.cfg) = False
INVENTORY_ENABLED(/home/cathal/workspace/ansible/ansible.cfg) = ['ini', 'yaml']
```

##### OS / ENVIRONMENT
[...]

PR #264, opened 2021-06-14
Fix current member comparison for correct sid

dihmandrake:
The value "$current_member.sid" is empty and therefore, the account is always added again to the group. As a result, the module reports changed every time.

This patch sets the correct sid for comparison.

The bug was likely introduced by 72e4a1acc73f92037b118557159854ece506b832 and fetching only the objectSID, which was not used for comparison.

Also, I think this will fix issue: https://github.com/ansible-collections/community.windows/issues/256


jborean93 (approved):
Thanks for fixing this so quickly, I've just added a commit that adds the changelog fragment so it shows up on the next release for `community.windows`.

PR #268, opened 2021-07-05
Added Conditional to check if disk should be brought online.

bateskevin:
##### SUMMARY

This fixes a bug where if you want to bring a disk offline by using the parameter "online" set to "false/no" it would bring the disk online again anyways. All needed is the added conditional to fix this, where the parameter gets taken in to account at the end.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_initialize_disk

###### Code used
```
- name: Take disk offline bases on UUID
      win_initialize_disk:
        uniqueid: "DiskUUID"
        online: no
        force: yes
```

###### Expected behaviour
We expected the Disk to be taken offline (and staying offline) after the ansible run.

###### Actual behaviour
The disk was shortly taken offline, but then brought back online by the line we fixed in the module.

### DIFF
--- a/changelogs/fragments/win_initialize_disk-offline.yml
+++ b/changelogs/fragments/win_initialize_disk-offline.yml
@@ -1,2 +1,3 @@
 bugfixes:
-- win_initialize_disk - Ensure ``online: False`` doesn't bring the disk online again - https://github.com/ansible-collections/community.windows/pull/268
+- >-
+  win_initialize_disk - Ensure ``online: False`` doesn't bring the disk online again - https://github.com/ansible-collections/community.windows/pull/268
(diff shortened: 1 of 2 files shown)


jborean93:
Doesn't the change here mean the disk is no longer initialised? Wouldn't the better method be to just take the disk back offline once it is initialized?

bateskevin:
Hi @jborean93,

Thanks for your reply! 

But in the case of $bring_online being set to false we do not want it to be online. Is there maybe a misunderstanding on our side here?

In the Ansible Playbook we specifically set 'online' to 'no'/'false'. This gets ignored without the conditional we commited.

Kind regards,
Kevin

jborean93 (approved):

jborean93:
My apologies (and sorry for the delay in replying). After thinking about it before my concerns doesn't seem to be valid. I've pushed a commit that adds a changelog fragment but everything else looks good to me.

PR #277, opened 2021-07-28
Fixed module param in win_dns_record.ps1

### LINKED ISSUE #276: Specifing a remote DNS server with computer_name doesn't work anymore since v1.6.0
##### SUMMARY
Specifying a remote computer_name in the `win_dns_record` module doesn't work anymore.

This is likely due the change in https://github.com/ansible-collections/community.windows/commit/f677f94d4126fa17ccab96a2dd50ed9cfbdbaa31#diff-795bc7066e5c113a02653007e502feb512eb54d1089b4dc95f73a316e70eda79R34 which changed the "module.Params.computer_name" to "module.Params.server" 

When you try to specify the module param server the role breaks. 

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
win_dns_record

##### ANSIBLE VERSION
2.10.9

##### COLLECTION VERSION
1.6.0

##### OS / ENVIRONMENT
Windows Server 2016

##### STEPS TO REPRODUCE
Create the following step to create DNS entries on a remote host using a jump windows host.
```yml
- name: Create Server DNS entries
  become: yes
  win_dns_record:
    computer_name: "{{ dns_server }}"
    name: "test-record"
    type: "A"
    value: "{{ host_ip }}"
    state: present
    zone: "example.com"
  delegate_to: ansible02.example.com
  vars:
    ansible_connection: winrm
    ansible_port: 5986
    ansible_winrm_transport: kerberos
    ansible_winrm_server_cert_validation: ignore
    ansible_user:
    ansible_password:
    ansible_become_method: runas
    ansible_become_user:
    ansible_become_pass:
```

##### EXPECTED RESULTS
The DNS entry is not created and the module fails with an error.

##### ACTUAL RESULTS
The DNS record is not created.

ntimo:
##### SUMMARY
This mr changes the `$module.Params.server` back to `$module.Params.computer_name` since server is not recognized in the module as an accepted parameter. And computer_name has been used before. If you want to change to server. I would suggest to document this correctly as a breaking change :)

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_dns_record

Fixes #276

### DIFF
--- a/changelogs/fragments/win_dns_record-comp-name.yml
+++ b/changelogs/fragments/win_dns_record-comp-name.yml
@@ -1,0 +1,2 @@
+bugfixes:
+- win_dns_record - Fix issue when trying to use the ``computer_name`` option - https://github.com/ansible-collections/community.windows/issues/276
--- a/plugins/modules/win_dns_record.ps1
+++ b/plugins/modules/win_dns_record.ps1
@@ -31,7 +31,7 @@ $type = $module.Params.type
 $values = $module.Params.value
 $weight = $module.Params.weight
 $zone = $module.Params.zone
-$dns_computer_name = $module.Params.server
+$dns_computer_name = $module.Params.computer_name
 $extra_args = @{}
 if ($null -ne $dns_computer_name) {
     $extra_args.ComputerName = $dns_computer_name


jborean93:
Thanks for the pickup, I've pushed a commit that adds a changelog fragment as well. Why this was changed is just a missed error in the review process. I don't think there are any plans on changing the name. If there was we would use an alias first to not make a breaking change.

PR #278, opened 2021-07-30
New Module: win_domain_ou take 2

gamethis:
The win_domain_ou module manages Active Directory Organizational Units (OUs).
ISSUE TYPE

    New Module Pull Request

COMPONENT NAME

win_domain_ou
ADDITIONAL INFORMATION


plugins/modules/win_domain_ou.py
```diff
@@ -0,0 +1,151 @@
...
+
+# Copyright: (c) 2020 VMware, Inc. All Rights Reserved.
+# SPDX-License-Identifier: GPL-3.0-only
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+DOCUMENTATION = r'''
+---
+module: win_domain_ou
+short_description: Manage Active Directory Organizational Units
+author: Joe Zollo (@joezollo), Larry Lane (@gamethis)
```
jborean93, line 12: Might be best to set this as a list. You also need to add `version_added: 1.7.0` to denote when this was added

```suggestion
version_added: 1.7.0
author:
- Joe Zollo (@joezollo)
- Larry Lane (@gamethis)
```

plugins/modules/win_domain_ou.py
```diff
@@ -0,0 +1,151 @@
...
+    description:
+    - The password for the domain you are accessing
+  filter:
+    type: str
+    description: filter for lookup of ou.
+    default: '*'
+  properties:
+    type: dict
+    description:
+      - Defines specific LDAP properties for the organizational unit.
```
jborean93, line 80: You should describe this a bit more. Currently they are LDAP properties but it looks like the code is trying to convert the values to and from snake and camel case. Personally I think if this is going to be a free form field that it just use the camel case names that are in use with LDAP. It makes it easier for people wanting to set the raw values outside of the root module options without having to translate what they are actually for.

jborean93 (changes requested):
Most of the information here looks good. I do have some concerns with trying to convert the LDAP properties to and from snake_cake. I personally think they should just be as is and have people set the actual LDAP property name in that dict and avoid all the conversions in the module. For example I think `postal_code` should stay as `postalCode` to better reflect the raw nature of `properties`.

What are your thoughts on this?

gamethis:
> Most of the information here looks good. I do have some concerns with trying to convert the LDAP properties to and from snake_cake. I personally think they should just be as is and have people set the actual LDAP property name in that dict and avoid all the conversions in the module. For example I think `postal_code` should stay as `postalCode` to better reflect the raw nature of `properties`.
> 
> What are your thoughts on this?

@jborean93 I was doing that just to try to follow standard ansible fact idea... but yes I actually prefer the other way where it is just a straight up property passthrough.  I will update make the update for this

jborean93:
>  I was doing that just to try to follow standard ansible fact idea.

I think if it was root options then using the snake_case is best but in this case it's meant to reflect the raw LDAP properties and reduces confusion so I'm totally fine having it without the translation.

> My question is do we return free form or does it need to be snake_case return of values?

Personally I don't think it should be returning values at all and this should be done under a `win_domain_ou_info` module but my thoughts are the return value should match the input options as closely as it can.

gamethis:
@jborean93 jborean93 so everything is fixed and working... the server 2016 test failed to ever connect with winrm from beginning... is there a way to rerun that?

gamethis:
> Most of the information here looks good. I do have some concerns with trying to convert the LDAP properties to and from snake_cake. I personally think they should just be as is and have people set the actual LDAP property name in that dict and avoid all the conversions in the module. For example I think `postal_code` should stay as `postalCode` to better reflect the raw nature of `properties`.
> 
> What are your thoughts on this?
Made the changes

plugins/modules/win_domain_ou.ps1
```diff
@@ -0,0 +1,242 @@
+#!powershell
+
+# Copyright: (c) 2020 VMware, Inc. All Rights Reserved.
+# SPDX-License-Identifier: GPL-3.0-only
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+#AnsibleRequires -CSharpUtil Ansible.Basic
+#Requires -Module Ansible.ModuleUtils.CamelConversion
```
jborean93, line 8: I don't think this is needed anymore right?

plugins/modules/win_domain_ou.ps1
```diff
@@ -0,0 +1,242 @@
...
+
+Function Get-OuObject {
+    Param([PSObject]$Object)
+    $obj = $Object | Select-Object -Property * -ExcludeProperty nTSecurityDescriptor | ConvertTo-Json -Depth 1 | ConvertFrom-Json
+    return $obj
+}
+
+# attempt import of module
+Try { Import-Module ActiveDirectory }
+Catch { $module.FailJson("Line 124: The ActiveDirectory module failed to load properly: $($_.Exception.Message)", $_) }
```
jborean93, line 124: Probably best to exclude `Line 124` in the error. `FailJson` will provide the stacktrace which includes the line numbers when you provide the error/exception which you have here. Ditto for the rest.

plugins/modules/win_domain_ou.py
```diff
@@ -0,0 +1,152 @@
...
+    description:
+    - The password for the domain you are accessing
+  filter:
+    type: str
+    description: filter for lookup of ou.
+    default: '*'
+  properties:
+    type: dict
+    description:
+      - Free form dict of properties for the organizational unit.
```
jborean93, line 81: It might be best to mention that these follow the `LDAP` property names, like `streetAddress`.

jborean93 (commented):
Just a few minor nits, looks like the doc examples also need to be updated to use the actual LDAP property names rather then the `snake_case` ones that were originally in place.

plugins/modules/win_domain_ou.ps1
```diff
@@ -0,0 +1,241 @@
...
+if ($module.Params.properties.count -ne 0){
+    $module.Params.properties.Keys | ForEach-Object{
+        $params.Add($_,$module.Params.properties.Item($_))
+    }
+}
+# capture settings to know what we are doing
+$module.Result.params = $params
+$module.Result.extra_args = $extra_args
+$module.Result.onboard_extra_args = $onboard_extra_args
+$module.Result.properties = $properties
```
jborean93, line 73: You should avoid doing this as it's a bit superfluous and could expose sensitive information. The `params` are automatically displayed in Ansible when you run with `-vvv` solving that problem and the extras are really just based on those values anyway so the data isn't really needed.

```suggestion
```

plugins/modules/win_domain_ou.py
```diff
@@ -0,0 +1,152 @@
...
+
+- name: Ensure OU is present with specific properties
+  community.windows.win_domain_ou:
+    name: WS1Users
+    path: "CN=EUC Users,DC=euc,DC=vmware,DC=lan"
+    protected: true
+    properties:
+      city: Sandy Springs
+      state: Georgia
+      street_address: 1155 Perimeter Center West
```
jborean93, line 113: Needs to be updated to use the LDAP form `streetAddress`.

plugins/modules/win_domain_ou.py
```diff
@@ -0,0 +1,152 @@
...
+      country: US
+      description: EUC Business Unit
+      postal_code: 30189
```
jborean93, line 116: Needs to be updated to use the LDAP form `postalCode`.

plugins/modules/win_domain_ou.py
```diff
@@ -0,0 +1,152 @@
...
+
+- name: Ensure OU updated with new properties
+  community.windows.win_domain_ou:
+    name: WS1Users
+    path: DC=euc,DC=vmware,DC=lan
+    protected: false
+    properties:
+      city: Atlanta
+      state: Georgia
+      managed_by: jzollo@vmware.com
```
jborean93, line 127: Needs to be updated to use the LDAP form `managedBy`.

plugins/modules/win_domain_ou.py
```diff
@@ -0,0 +1,152 @@
...
+    modified:
+    protected:
+    properties:
+      display_name:
+      description:
+      city:
+      street_address:
+      postal_code:
+      country:
+      managed_by:
```
jborean93, line 151: These also need fixing to reflect that we don't convert to snake_case anymore.

jborean93 (approved):

jborean93:
I've unfortunately had to revert this commit as further testing found some larger problems https://github.com/ansible-collections/community.windows/pull/297. It seems like some of the tests have been written around actual results rather than the proper expectations, testing something had not changed when in reality it should for that scenario.

gamethis:
I don't fully comprehend the difference in the final test and why it shows different from the other... bu t I think I have found the error in the code that caused it... I will go back and iprove the test and reoopen a pr in next day or so

PR #280, opened 2021-08-10
Module win_route upgraded

iVcente:
##### SUMMARY

- Possibility of adding a static route with the same network destination and gateway, but different mask.
- Moving from `New-NetRoute` and `Remove-NetRoute` PowerShell commands to `route add` and `route delete` CMD commands. The PowerShell commands don't work as expected, while the CMD ones do the job just right.

The idea of adding a `gateway` parameter as a requirement was to help preventing any errors by adding something wrong and causing a big trouble. I mean, why not add a gateway? Even if you actually need a `0.0.0.0` one, it's good to be explicit about it.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

win_route

##### ADDITIONAL INFORMATION

I had to create a playbook to add/remove static routes on servers in the company I work for and when I verified the results achieved with the current module, there were some little problems that stopped us from using it. Later, I decided to give it a shot and try to update the module.

Now, it is possible to add (or remove) routes with same network destination and gateway, but with a different mask! Unfortunately, in order to do that, the utilisation of CIDR notation had to be sacrificed.

Also, we faced some problems with the PowerShell commands utilised to add and remove routes: `New-NetRoute` and `Remove-NetRoute`. Although Windows successfully executes these commands - and shows the route on its routing table -, for some reason they don't really work as intended in practice. Servers wouldn't communicate with each other after adding a static route by these commands/module. The solution was to use the old but gold CMD commands: `route add` and `route delete`. After these little fixes, this modified version is currently being used in our production environment with a 100% of success so far.
Doing a little bit of research I found some people with the exactly same problem with these PowerShell commands, the solution to everyone was to abandon the PowerShell commands in favor of the CMD ones.

So after this trouble, I thought about sharing these changes with the community. These fixes helped us a lot and may help others as well. This is my first PR, so in case of any issues, just let me know - and sorry by them.

The documentation was updated as well.

### DIFF
--- a/plugins/modules/win_route.ps1
+++ b/plugins/modules/win_route.ps1
@@ -1,6 +1,7 @@
 #!powershell
 
 # Copyright: (c) 2016, Daniele Lazzari <lazzari@mailup.com>
+# Modified by: Vicente Danzmann Vivian (@iVcente)
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
 #Requires -Module Ansible.ModuleUtils.Legacy
@@ -10,8 +11,9 @@
 $params = Parse-Args $args -supports_check_mode $true
 
 $check_mode = Get-AnsibleParam -obj $params -name "_ansible_check_mode" -default $false
-$dest = Get-AnsibleParam -obj $params -name "destination" -type "str" -failifempty $true
-$gateway = Get-AnsibleParam -obj $params -name "gateway" -type "str"
+$network = Get-AnsibleParam -obj $params -name "network" -type "str" -failifempty $true
+$mask = Get-AnsibleParam -obj $params -name "mask" -type "str" -failifempty $true
+$gateway = Get-AnsibleParam -obj $params -name "gateway" -type "str" -failifempty $true
 $metric = Get-AnsibleParam -obj $params -name "metric" -type "int" -default 1
 $state = Get-AnsibleParam -obj $params -name "state" -type "str" -default "present" -validateSet "present","absent"
 $result = @{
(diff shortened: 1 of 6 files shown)


plugins/modules/win_route.py
```diff
@@ -11,16 +12,21 @@
 description:
     - Add or remove a static route.
 options:
-  destination:
+  network:
```
jborean93, line 15: You cannot change a module option name as that will break existing users of the module. If `network` is a desired name you will either have to make it an alias for `destination` or vice versa. This allows people to continue to use the older option as well as the newer one at the same time.

plugins/modules/win_route.py
```diff
@@ -11,16 +12,21 @@
...
+    description:
+      - Destination network's mask.
     type: str
     required: yes
   gateway:
     description:
         - The gateway used by the static route.
-        - If C(gateway) is not provided it will be set to C(0.0.0.0).
     type: str
+    required: yes
```
jborean93, line 29: This will break existing users who don't set `gateway` and rely on the defaults.

plugins/modules/win_route.py
```diff
@@ -37,26 +43,30 @@
...
+    gateway: 192.168.24.1
     state: absent
 '''
 RETURN = r'''
 output:
     description: A message describing the task result.
     returned: always
     type: str
-    sample: "Route added"
+    sample: "Route added!"
```
jborean93, line 71: You shouldn't change the output value unless there's a good reason to. You never know what people might be parsing/expecting in their playbooks.

plugins/modules/win_route.ps1
```diff
@@ -10,95 +11,120 @@
...
+        $result.output = "Route can be removed!"
+        return
+    }
 
-  Add-Route -Destination $dest -Gateway $gateway -Metric $metric -CheckMode $check_mode
+    # If route exists, remove it
+    if ($Route) {
+        try {
+            # Remove static route
+            route DELETE $Network MASK $Mask $Gateway
```
jborean93, line 108: The trouble with moving to a native command from a cmdlet is that error processing is a whole different beast with PowerShell. There are so many different behaviour permutations when it comes to detecting errors and formatting the output that I'm never confident this will actually catch an error.

jborean93 (changes requested):
There are a few changes in here that will break people who currently use the module. You will need to ensure that any new additions don't break things that work today or if they do there's a really good reason for that.

> Now, it is possible to add (or remove) routes with same network destination and gateway, but with a different mask! Unfortunately, in order to do that, the utilisation of CIDR notation had to be sacrificed.

That shouldn't be a hurdle. It should be possible to parse the value, extract the address length and derive the netmask from there.

> Although Windows successfully executes these commands - and shows the route on its routing table -, for some reason they don't really work as intended in practice.

While I can't claim to understand any of these components I highly recommend avoiding native calls over PowerShell cmdlets. There has to be some difference in the PowerShell method where hopefully it requires a small tweak to fix up. The problem with calling a native binary is dealing with errors. The behaviour in PowerShell when it comes to dealing with errors is quite confusing and sometimes are just missed altogether.

If you still need to use the `route` command then calling it through our [Start-AnsibleWindowsProcess](https://github.com/ansible-collections/ansible.windows/blob/main/plugins/module_utils/Process.psm1#L119) module util is highly recommended. Essentially you want to add `#AnsibleRequires -PowerShell ansible_collections.ansible.windows.plugins.module_utils.Process` to the top of the script then do the following

```powershell
$res = Start-AnsibleWindowsProcess -FilePath route.exe -ArgumentList @('DELETE', $Network, 'MASK', $Mask, $Gateway)
if ($res.ExitCode -ne 0) {
    $result.stdout = $res.Stdout
    $result.stderr = $res.Stderr
    $result.rc = $res.ExitCode
    Fail-Json $result "Failed to remove route, see stdout/stderr for more details"
}
```

This way you are ensuring you are failing if the rc was not 0 and you've definitely captured/returning the stdout/stderr from that process back to the user.

PR #283, opened 2021-08-15
win_lineinfile: fixes the inconsistent of the last emtpy line on diff.

teramako:
##### SUMMARY

The last empty line is stripped from the only before data of diff.
While the after data of diff is fixed on #219.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

win_lineinfile

##### ADDITIONAL INFORMATION

Related #219

```paste below

```


jborean93:
Thanks for the bugfix, I've pushed a commit that added the changelog fragment.

PR #289, opened 2021-08-26
win_domain_user: Fix issue when trying to update user password when u…

dev-ddoe:
win_domain_user: Use SAM when user have no upn during password verification

##### SUMMARY
On AD, explicit upn is not mandatory.

When the following playbook is played
```  community.windows.win_domain_user:
    name: bob
    firstname: Bob
    surname: Smith
    password: B0bP4ssw0rd
    update_password: when_changed
    state: present
    domain_username: DOMAIN\admin-account
    domain_password: SomePas2w0rd
    domain_server: domain@DOMAIN.COM
```

it works properly the first time but when it is played a second time,
it fails because password verification is done using upn of the user.
But user does not have a upn

The error returned is
```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was:    at CallSite.Target(Closure , CallSite , Type , String , String , String , String , String )
fatal: [ad1]: FAILED! => {"changed": false, "msg": "Unhandled exception while executing module: Failed to logon  (The parameter is incorrect, Win32ErrorCode 87 - 0x00000057)"}
```

##### WORKAROUND

using
```
update_password: always
```
prevents password verification

### DIFF
--- a/changelogs/fragments/win_domain_user-missing-upn.yml
+++ b/changelogs/fragments/win_domain_user-missing-upn.yml
@@ -1,0 +1,2 @@
+bugfixes:
+- win_domain_user - Fallback to NETBIOS username for password verification check if the UPN is not set - https://github.com/ansible-collections/community.windows/pull/289
--- a/plugins/modules/win_domain_user.ps1
+++ b/plugins/modules/win_domain_user.ps1
@@ -186,12 +186,14 @@ If ($state -eq 'present') {
         If ($new_user -or ($update_password -eq "always")) {
             $set_new_credentials = $true
         } elseif ($update_password -eq "when_changed") {
-            If ($user_obj.UserPrincipalName) {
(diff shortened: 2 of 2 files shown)


jborean93:
Is there a way to get the NETBIOS domain part from the AD user object? My only concern with this is if the `sAMAccountName` is also the same name as a local user on the host in question. Being able to do `DOMAIN\username` as the fallback if the UPN is not present should solve that problem.

dev-ddoe:
Good point. 
I've updated the PR in order to get the constructed attribute **msDS-PrincipalName** which contain the username in the form:
DOMAIN\user

jborean93 (approved):
Thanks for the PR, I pushed a slight change which added the changelog fragment and a tweak so that `Test-Credential` is only called in 1 place rather than be doubled up.

dev-ddoe:
It looks good. thank you.

PR #294, opened 2021-09-21
Added Server 2022 to CI matrix

jborean93:
##### SUMMARY
Adds Server 2022 to testing matrix.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
tests


jborean93:
Failures on group 4 seem to be down to a change in `ansible-test` on devel with the controller/node split. Will have to investigate it further but as Server 2022 passes on the other groups this change is fine.

PR #298, opened 2021-09-23
Fix win_firewall bools comparation

lydiym:
##### SUMMARY
win_firewall fix for windows versions compatibility increasing

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_firewall

##### ADDITIONAL INFORMATION
The code below works in `ps/wmf 5.x+` but it is wrong way to compare two different types(`GpoBoolean`/`Boolean`). And also makes unjustified incompatible with older versions(in `win8.1`/`srv12r2` result is `False` with all the consequences :( )
```powershell
[Microsoft.PowerShell.Cmdletization.GeneratedTypes.NetSecurity.GpoBoolean]::True -eq $true
```


jborean93 (approved):
Thanks for the pickup!

PR #299, opened 2021-09-24
New Module: win_domain_ou

gamethis:
he win_domain_ou module manages Active Directory Organizational Units (OUs).
ISSUE TYPE

New Module Pull Request

COMPONENT NAME

win_domain_ou
ADDITIONAL INFORMATION

fixed from #278


gamethis:
@jborean93 reworked the test and module.  Should be good to go now.

plugins/modules/win_domain_ou.ps1
```diff
@@ -0,0 +1,257 @@
...
+}Catch{$module.FailJson("Get-ADOrganizationalUnit failed: $($_.Exception.Message)", $_) }
+
+# set path if not defined to base domain
+if ($null -eq $path){
+    if ($($all_ous | Measure-Object | Select-Object -ExpandProperty Count) -eq 1){
+        $matched = $all_ous.DistinguishedName -match "DC=.+"
+    }elseif ($($all_ous | Measure-Object | Select-Object -ExpandProperty Count) -gt 1) {
+        $matched = $all_ous[0].DistinguishedName -match "DC=.+"
+    }else{
+        $module.FailJson("Path was null and unable to determine default domain $($_.Exception.Message)", $_)
```
jborean93, line 154: There's no error record/exception so the 2nd argument isn't usable here

```suggestion
        $module.FailJson("Path was null and unable to determine default domain $($_.Exception.Message)")
```

plugins/modules/win_domain_ou.ps1
```diff
@@ -0,0 +1,257 @@
...
+    }
+    if ($matched){
+        $path = $matches.Values[0]
+    }else{
+        $module.FailJson("Unable to find default domain $($_.Exception.Message)", $_)
```
jborean93, line 159:
```suggestion
        $module.FailJson("Unable to find default domain $($_.Exception.Message)")
```

plugins/modules/win_domain_ou.ps1
```diff
@@ -0,0 +1,257 @@
...
+    }
+}
+$module.Result.path = $path
```
jborean93, line 162: This needs to be document in the module docs, currently there's only `ou`.

plugins/modules/win_domain_ou.ps1
```diff
@@ -0,0 +1,257 @@
...
+        Try {
+            New-ADOrganizationalUnit @params @onboard_extra_args -ProtectedFromAccidentalDeletion $protected -WhatIf:$check_mode
+        }Catch {
+            $module.FailJson("Failed to create organizational unit: $($_.Exception.Message)", $_)
+        }
+    }
+    # ou exists, update object
+    if ($current_ou) {
+        Try {
+            Set-ADOrganizationalUnit -Identity "OU=$name,$path" @params @onboard_extra_args -WhatIf:$check_mode
```
jborean93, line 189: There are no checks here to determine whether this call is actually required or not.

plugins/modules/win_domain_ou.ps1
```diff
@@ -0,0 +1,257 @@
...
+} Catch {
+    $module.Diff.before = $null
+    $current_ou = $null
+}
+if ($state -eq "present") {
+    # ou does not exist, create object
+    if(-not $current_ou) {
+        $params.Name = $name
+        $params.Path = $path
+        $module.Result.params = $params
```
jborean93, line 179: This seems like a leftover when debugging, this shouldn't be returned.

plugins/modules/win_domain_ou.ps1
```diff
@@ -0,0 +1,257 @@
...
+            Set-ADOrganizationalUnit -Identity "OU=$name,$path" -ProtectedFromAccidentalDeletion $false @onboard_extra_args -Confirm:$False -WhatIf:$check_mode
+        }Catch{
+            $module.FailJson("Failed to remove ProtectedFromAccidentalDeletion Lock: $($_.Exception.Message)", $_)
+        }
+            # check recursive deletion
+        if ($recursive) {
+            try{
+                Remove-ADOrganizationalUnit -Identity "OU=$name,$path" -Confirm:$False -WhatIf:$check_mode -Recursive @onboard_extra_args
+            }catch{
+                $module.FailJson("Failed to recurssively Remove-ADOrganizationalUnit $($_.Exception.Message)", $_)
```
jborean93, line 210:
```suggestion
                $module.FailJson("Failed to recursively Remove-ADOrganizationalUnit $($_.Exception.Message)", $_)
```

plugins/modules/win_domain_ou.ps1
```diff
@@ -0,0 +1,257 @@
...
+            try{
+                Remove-ADOrganizationalUnit -Identity "OU=$name,$path" -Confirm:$False -WhatIf:$check_mode @onboard_extra_args
+            }Catch{
+                $module.FailJson("Failed to Remove-ADOrganizationalUnit: $($_.Exception.Message)", $_)
+            }
+        }
+    }
+}
+
+# determine if a change was made
```
jborean93, line 222: This shouldn't be needed at all, `$module.Result.changed = $true` should be set after every operation where a change will occur (where you call `New/Set/Remove-ADOrganizationalUnit`. The same applies with `$module.Diff.after` you should be adding that in the code where the changes are being applied rather than at the end here.

The biggest benefits is now you don't need to split the change detection logic for when you are in check mode and not and the diff could potentially be simplified in some cases (absent will just always set after as `""`.

plugins/modules/win_domain_ou.py
```diff
@@ -0,0 +1,164 @@
...
+    description:
+    - The password for the domain you are accessing
+  filter:
+    type: str
+    description: filter for lookup of ou.
+    default: '*'
+  properties:
+    type: dict
+    description:
+      - Free form dict of properties for the organizational unit. Follows LDAP property names, like StreetAddress or PostalCode.
```
jborean93, line 81: It's always nice to format some important entries with `C()`. Also a minor nit is that the properties are in `camelCase` not `PascalCase`.

```suggestion
      - Free form dict of properties for the organizational unit. Follows LDAP property names, like C(streetAddress) or C(postalCode).
```

plugins/modules/win_domain_ou.py
```diff
@@ -0,0 +1,164 @@
...
+      city: Atlanta
+      state: Georgia
+      managedBy: jzollo@vmware.com
+  delegate_to: win-ad1.euc.vmware.lab
+'''
+
+RETURN = r'''
+ou:
+  description: New/Updated organizational unit parameters
+  returned: When l(state=present)
```
jborean93, line 134: use `I()` to italicize text, `l()` is for URL links.

plugins/modules/win_domain_ou.py
```diff
@@ -0,0 +1,164 @@
...
+  type: dict
+  sample:
```
jborean93, line 136: Can you double check whether properties are returned in `camelCase` or `PascalCase`, I always assumed for former.

tests/integration/targets/win_domain_ou/tasks/check_mode_test.yml
```diff
@@ -0,0 +1,87 @@
+---
+- name: Run Check Mode Tests
+  block:
+    - name: Ensure OU is present (check_mode)
+      community.windows.win_domain_ou:
+        name: ansible_checkmode
+      register: test1c
+      failed_when: test1c is changed
```
jborean93, line 8: This should be `changed`. A module reports a change in check mode if it was going to make a change (even though it didn't). This leads into one of my comments above where the changed result should be set in the places you call a cmdlet that makes a change rather than do it on the result.

tests/integration/targets/win_domain_ou/tasks/check_mode_test.yml
```diff
@@ -0,0 +1,87 @@
...
+        path: "{{ win_domain_ou_root_path }}"
+        properties:
+          city: Sandy Springs
+          state: Georgia
+          streetaddress: 1155 Perimeter Center West
+          country: US
+          description: EUC Business Unit
+          postalcode: 30189
+      register: test2c
+      failed_when: test2c is changed
```
jborean93, line 23: Same here, this should be changed if it updated properties.

jborean93 (changes requested):
Added some comments, the tests around detecting a change in check mode still seem to be wrong.

plugins/modules/win_domain_ou.py
gamethis: So vast majority of items come back in `PascalCase` ... a few come back `camelCase`

Sample output:
```shell
CanonicalName                   : ansible.test/VMware/End User Computing
City                            :
CN                              :
Country                         :
Created                         : 10/1/2021 5:05:44 PM
createTimeStamp                 : 10/1/2021 5:05:44 PM
Deleted                         :
Description                     :
DisplayName                     :
DistinguishedName               : OU=End User Computing,OU=VMware,DC=ansible,DC=test
dSCorePropagationData           : {1/1/1601 12:00:00 AM}
instanceType                    : 4
isDeleted                       :
LastKnownParent                 :
LinkedGroupPolicyObjects        : {}
ManagedBy                       :
Modified                        : 10/1/2021 5:05:44 PM
modifyTimeStamp                 : 10/1/2021 5:05:44 PM
Name                            : End User Computing
nTSecurityDescriptor            : System.DirectoryServices.ActiveDirectorySecurity
ObjectCategory                  : CN=Organizational-Unit,CN=Schema,CN=Configuration,DC=ansible,DC=test
ObjectClass                     : organizationalUnit
ObjectGUID                      : f6b8a10c-e5c5-4be8-9a89-1233cb1cad8f
ou                              : {End User Computing}
PostalCode                      :
ProtectedFromAccidentalDeletion : False
sDRightsEffective               : 15
State                           :
StreetAddress                   :
uSNChanged                      : 16501
uSNCreated                      : 16501
whenChanged                     : 10/1/2021 5:05:44 PM
whenCreated                     : 10/1/2021 5:05:44 PM
```

plugins/modules/win_domain_ou.ps1
gamethis: agreed removing

plugins/modules/win_domain_ou.ps1
gamethis: added to document

plugins/modules/win_domain_ou.py
gamethis: it is perdominately PasalCase with a few camelCase thrown in

plugins/modules/win_domain_ou.py
gamethis: updated

plugins/modules/win_domain_ou.ps1
gamethis: added checks

tests/integration/targets/win_domain_ou/tasks/check_mode_test.yml
gamethis: Thank you misunderstood how this works updated and fixed

plugins/modules/win_domain_ou.ps1
gamethis: updated code to reflect what you said.  it did simplify much of the code.

gamethis:
@jborean93 thanks for the reviews on this module.  I believe it is up to par with the requests that have been made.

Is there anything else still lacking?

PR #307, opened 2021-10-06
win_scheduled_task: Require password only for password logon type

patrickcjh:
##### SUMMARY
When using logon_type of s4u, password should not be required. Providing a password will cause the task registration to fail.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_scheduled_task.ps1

### DIFF
--- a/plugins/modules/win_scheduled_task.ps1
+++ b/plugins/modules/win_scheduled_task.ps1
@@ -705,8 +705,8 @@ if ($null -ne $username -and $null -ne $group) {
     Fail-Json -obj $result -message "username and group can not be set at the same time"
 }
 if ($null -ne $logon_type) {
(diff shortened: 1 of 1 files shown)


jborean93:
Looks like there's a test expectation for this that needs to be fixed https://github.com/ansible-collections/community.windows/blob/48dfbf5354131f65c353d852899b55856005be6b/tests/integration/targets/win_scheduled_task/tasks/failures.yml#L31-L37. It would be great if you could add an S4U task logon type test to verify the task was created as well.

It seems like the existing error check is for when the S4U principal is not the current user (username is provided). Currently only that scenario works when both the username and password is supplied but it may in fact be a bug in the code as you should still be able to specify the S4U principal without it's password. If you feel up to it you could try and investigate where that bug is and potentially fix it.

PR #325, opened 2021-10-27
New Module win_domain_computer_info

gamethis:
##### SUMMARY

module to allow users to query for domain computer information from active directory.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME
win_domain_computer_info

##### ADDITIONAL INFORMATION
sample of how module will work
```
- name: Get info for current computer
  community.windows.win_domain_computer_info:
    name: "{{ ansible_host }}"
    register: computer_info

- name: Get Computer Info with Properties
  community.windows.win_domain_computer_info:
    name: "{{ ansible_hostname }}"
    properties:
      - name
      - samaccountname
      - dnshostname
  register: computer_info_with_properties
```


plugins/modules/win_domain_computer_info.ps1
```diff
@@ -0,0 +1,98 @@
+#!powershell
+Set-StrictMode -Version 2.0
```
jborean93, line 2: Just for the future this isn't needed unless you really want it. I personally find strict mode a bit too strict and fights against standard pwsh-isms but it's ultimately up to you the module author what you want to do.

plugins/modules/win_domain_computer_info.ps1
```diff
@@ -0,0 +1,98 @@
...
+
+#AnsibleRequires -CSharpUtil Ansible.Basic
+#Requires -Module ActiveDirectory
+
+$spec = @{
+    options = @{
+        identity = @{ type = "str";  }
+        properties = @{type = "list"; elements = "str"; default = @('*')}
+        excluded_properties = @{type= "list"; elements = "str"; default = @('DeclaredConstructors', 'DeclaredMembers', 'nTSecurityDescriptor')}
+        search_scope = @{ type = 'str'; choices = @('Base', 'OneLevel', 'Subtree')}
```
jborean93, line 15: Have a look at `win_domain_object_info` https://github.com/ansible-collections/community.windows/blob/main/plugins/modules/win_domain_object_info.ps1#L20 to see how this is done there.

plugins/modules/win_domain_computer_info.py
```diff
@@ -0,0 +1,197 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2020, Ansible Project
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+DOCUMENTATION = r'''
+---
+module: win_domain_computer_info
+short_description: Gather active diretory information about a computer.
```
jborean93, line 10: directory*

plugins/modules/win_domain_computer_info.py
```diff
@@ -0,0 +1,197 @@
...
+  - Larry Lane (@gamethis)
+'''
+
+EXAMPLES = r'''
+- name: Install ActiveDirectory Powershell extensions
+  ansible.windows.win_feature:
+    name: RSAT-AD-PowerShell
+    state: present
+
+- name: Install Active Directory Module
```
jborean93, line 85: Does this actually work? I thought it can only be installed from the Windows feature. I also cannot find any module called `ActiveDirectory` on the PowerShell Gallery.

jborean93 (commented):
I've only done a very brief review but I'm not sure if this is really something that's needed considering we already have [win_domain_object_info](https://docs.ansible.com/ansible/latest/collections/community/windows/win_domain_object_info_module.html). Is there anything this module does that cannot already be achieved with `win_domain_object_info`?

plugins/modules/win_domain_computer_info.py
gamethis: Yes it does... The active directory module is installed when you install RSAT-AD-PowerShell which is located on the server. it does not come from the PowerShell Gallery

jborean93: This step is using `win_psmodule` to install the module which is what I was referring to. Because it comes from `RSAT` then the step above is what is needed and this particular example isn't right AFAIK.

PR #330, opened 2021-11-10
fix for #316

### LINKED ISSUE #316: win_domain_user created account could not be enabled
##### SUMMARY
Running win_domain_user to create a new account failed, becaus of password policy violation. Running this job next time will not enable this accout although the passwor is valid and does not violate the policy.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
community.windows.win_domain_user

##### ANSIBLE VERSION
```
2.11.4
```

##### COLLECTION VERSION
```
community.windows             1.6.0
```

##### CONFIGURATION
```
nothing changed
```

##### OS / ENVIRONMENT
Ubuntu 20.04 as ansible controller and also produceable with AWX 19.3.0 

##### STEPS TO REPRODUCE
```yaml
- community.windows.win_domain_user:
        name: Bob
        password: 123
        upn: Bob@mydomain
        state: present
        update_password: when_changed
        password_never_expires: true
        account_locked: no
        enabled: yes
```
Rerun this step with a valid password
```yaml
- community.windows.win_domain_user:
        name: Bob
        password: B0bP4ssw0rd#
        upn: Bob@mydomain
        state: present
        update_password: when_changed
        password_never_expires: true
        account_locked: no
        enabled: yes
```

##### EXPECTED RESULTS

The user should be created and enabled with the correct version of the password.

##### ACTUAL RESULTS
The user account was created during first run. Because of an empty passsword it is disabled. That's okay for security reason. At the second i get the following message:

```
FAILED! => {"changed": false, "msg": "Unhandled exception while executing module: Failed to logon Bob@mydomain (This user can't sign in because this account is currently disabled, Win32ErrorCode 1331 - 0x00000533)"
```

It works if i change the parameter "update_password" from  when_changed to always. But next run i will only change this password if it was really different. So i have not chance to enable an user account.

gamethis:
##### SUMMARY
This PR adds test and 
fixes #316 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_domain_user

##### ADDITIONAL INFORMATION
It adds the failed_test for the reported error.
Adds a Try catch around set-adpassword to give an appropriate message.
and Makes it so that the module appropriately appropriately returns created = True and changed = True on the error provided from #316


plugins/modules/win_domain_user.ps1
```diff
@@ -199,7 +200,11 @@ If ($state -eq 'present') {
         }
         If ($set_new_credentials) {
             $secure_password = ConvertTo-SecureString $password -AsPlainText -Force
-            Set-ADAccountPassword -Identity $user_guid -Reset:$true -Confirm:$false -NewPassword $secure_password -WhatIf:$check_mode @extra_args
+            try {
+                Set-ADAccountPassword -Identity $user_guid -Reset:$true -Confirm:$false -NewPassword $secure_password -WhatIf:$check_mode @extra_args
+            }catch{
+                Fail-Json $result "Failed to set password on account"
```
jborean93, line 206: If you are going to add the try/catch I would also set the exception message

```suggestion
               Fail-Json $result "Failed to set password on account: $($_.Exception.Message)"
```

Personally I think the try/catch isn't really that useful, the error message from Ansible should indicate the line where the failure occurred and considering the cmdlet is called `Set-ADAccountPassword` I feel it's pretty self explanatory.

jborean93 (approved):

PR #353, opened 2022-01-04
RST edit notification

briantist:
##### SUMMARY

After the WWG meeting, got inspired and whipped up this workflow that should notify PR authors that they modified the RST files directly and should edit the `.py` files instead.

Because this uses `pull_request_target`, it will need to be merged in `main` before it can be "tested" with a dummy PR.

This also doesn't remove the comment if the edits are removed; that would need a separate but similar workflow with the `paths:` inverted I think, or maybe some conditional stuff.

Also threw in a README in that docs dir, in case that deters someone from submitting the wrong PR in the first place.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

github workflows

##### ADDITIONAL INFORMATION

The RST files in `docs/` are auto-generated but people try to put up PRs for them sometimes to update the documentation.

### DIFF
--- a/.github/workflows/rst-pr.yml
+++ b/.github/workflows/rst-pr.yml
@@ -1,0 +1,35 @@
+name: RST PR
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
+    paths:
+      - "docs/*.rst"
+
+jobs:
+  notify:
+    permissions:
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


jborean93 (approved):
Thanks for the PR, hopefully it helps people who submit doc fix PRs in the future.

briantist:
@jborean93 would you like me to submit the same for `ansible.windows`?

jborean93:
Yea makes sense to keep them the same.

PR #355, opened 2022-01-10
[DO NOT MERGE] test pr for #353

briantist:
##### SUMMARY

testing #353 with an incorrect change to documentation

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

laps

### DIFF
--- a/docs/community.windows.laps_password_lookup.rst
+++ b/docs/community.windows.laps_password_lookup.rst
@@ -16,7 +16,7 @@ community.windows.laps_password
 
 Synopsis
 --------
-- This lookup returns the LAPS password set for a server from the Active Directory database.
+- This lookup returns the LAPS password set for a server from Active Directory.
 - See https://github.com/jborean93/ansible-lookup-laps_password for more information around installing pre-requisites and testing.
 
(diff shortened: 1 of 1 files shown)


jborean93:
Too quick for me, was just about to submit a PR :) The comment looks great and definitely better than before.

briantist:
@jborean93 perhaps the message could use some emojis or other attention grabbing 🛑⚠❌🚫

jborean93:
Personally I find them distracting but don't really mind either way. If anything it would be cool to substitute the `.py file(s)` with the actual link to the `.py` file but that's really just wishful thinking for the future.

briantist:
> Personally I find them distracting but don't really mind either way. If anything it would be cool to substitute the `.py file(s)` with the actual link to the `.py` file but that's really just wishful thinking for the future.

sgtm, I won't add any emojis then :)

The link would be nice, I'm not sure how easy it would be to generate the link accurately just from knowing the RST, like if there were a module and a plugin that shared the same name, or things like that. Probably doable, but will have to be a later contribution.

PR #362, opened 2022-02-02
Stabilise domain creation test tasks

jborean93:
##### SUMMARY
Tries to stabilize the domain creation tasks by waiting for ADWS to be online before continuing.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
setup_domain_tests

### DIFF
--- a/tests/integration/targets/setup_domain_tests/tasks/main.yml
+++ b/tests/integration/targets/setup_domain_tests/tasks/main.yml
@@ -4,17 +4,14 @@
     name: ansible-tester
   register: rename_host
 
-- name: Reboot
+- name: Reboot after changing hostname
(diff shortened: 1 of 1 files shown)


jborean93:
@briantist turns out Server 2019 was sometimes coming back from the reboot before the host was ready. The registry key it checked for was there so it was right to complete early. I have no idea why this only sometimes happens and I've never seen Windows create that key before the logon screen appeared. Ultimately for the sack of having stable tests I just added a dumb while loop that kept on trying to access ADWS for x amount of time. This waits until the logon screen appears like `win_reboot` should have and ADWS is then ready for the tests.

I ran this about 5 times and was no longer able to replicate the previous error.

PR #373, opened 2022-03-29
add new stable branch and fix up 2.14 santiy ignores

jborean93:
##### SUMMARY
Ansible has just bumped versions in the devel branch to 2.14 which requires a new sanity ignore file. This adds that file and ensures we still test with the newly created stable-2.13. Also fixes a bug in the RDS modules now that the controller provided util does SID lookups slightly differently when encountering a UPN formatted name.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ansible.windows

### DIFF
--- a/.azure-pipelines/azure-pipelines.yml
+++ b/.azure-pipelines/azure-pipelines.yml
@@ -54,6 +54,19 @@ stages:
               test: sanity
             - name: Units
               test: units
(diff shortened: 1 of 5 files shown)


jborean93:
Failures are unrelated and will be fixed in a separate PR.

PR #379, opened 2022-04-05
New module: win_listen_ports_facts

david-ns:
##### SUMMARY
New module win_listen_ports_facts to gather information about the ports of the machine

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
- win_listen_ports_facts

##### ADDITIONAL INFORMATION
- Recopilates the information of the TCP and UDP ports of the machine and the related processes.
- State of the TCP ports could be filtered, as well as the format of the date when the parent process was launched.
- The module's goal is to replicate the functionality of the linux module listen_ports_facts (community.general.listen_ports_facts), mantaining the format of the said module.

### DIFF
--- a/plugins/modules/win_listen_ports_facts.ps1
+++ b/plugins/modules/win_listen_ports_facts.ps1
@@ -1,0 +1,81 @@
+#!powershell
+
+# Copyright: (c) 2022, DataDope (@datadope-io)
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+#AnsibleRequires -CSharpUtil Ansible.Basic
+
+$spec = @{
+    options = @{
+        date_format = @{ type = 'str'; default = '%c' }
+        tcp_filter = @{ type = 'list'; elements = 'str'; default = 'Listen' }
+    }
+    supports_check_mode = $true
+}
+
+$module = [Ansible.Basic.AnsibleModule]::Create($args, $spec)
+
+$date_format = $module.Params.date_format
+$tcp_filter = $module.Params.tcp_filter
+
+# Structure of the response the script will return
+$ansibleFacts = @{
+    tcp_listen = @()
+    udp_listen = @()
+}
+
+# Build an index of the processes based on the PID
+$processes = @{}
+Get-Process -IncludeUserName | ForEach-Object {
+    $processes[$_.Id] = $_
+}
+
+# Format the given date with the same format as listen_port_facts stime (Date and time - abbreviated) by default, or
+# with the given format
+function Format-Date {
+    param (
+        $date
+    )
+
+    if ($null -ne $date) {
+        $date = Get-Date $date -UFormat $date_format
+    }
+
+    return $date
+}
+
+# Return the processed listener and the associated PID data
+function Build-Listener {
+    param (
+        $listener,
+        $type
+    )
+
+    return @{
+        address = $listener.LocalAddress
+        name = $processes[[int]$listener.OwningProcess].ProcessName
+        pid = $listener.OwningProcess
+        port = $listener.LocalPort
+        protocol = $type
+        stime = Format-Date $processes[[int]$listener.OwningProcess].StartTime
+        user = $processes[[int]$listener.OwningProcess].UserName
+    }
+}
+
+try {
+    # Retrieve the information of the TCP ports with Listen status by default, or with the given state/s
+    Get-NetTCPConnection -State $tcp_filter -ErrorAction SilentlyContinue | Foreach-Object {
+        $ansibleFacts.tcp_listen += Build-Listener $_ "tcp"
+    }
+
+    # Retrieve the information of the UDP ports
+    Get-NetUDPEndpoint | Foreach-Object {
+        $ansibleFacts.udp_listen += Build-Listener $_ "udp"
+    }
+}
+catch {
+    $module.FailJson("An error occurred while retrieving ports facts: $($_.Exception.Message)", $_)
+}
+
+$module.Result.ansible_facts = $ansibleFacts
+$module.ExitJson()
--- a/plugins/modules/win_listen_ports_facts.ps1
+++ b/plugins/modules/win_listen_ports_facts.ps1
@@ -26,8 +26,8 @@ $ansibleFacts = @{
 
 # Build an index of the processes based on the PID
 $processes = @{}
-Get-Process -IncludeUserName | ForEach-Object {
-    $processes[$_.Id] = $_
+Get-WmiObject -Class Win32_Process | ForEach-Object {
+    $processes[[int]$_.ProcessId] = $_
 }
 
 # Format the given date with the same format as listen_port_facts stime (Date and time - abbreviated) by default, or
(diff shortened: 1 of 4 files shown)


jborean93:
Without going too deep into the code the module seems quite decent and up to a nice standard. My concern is that `community.windows` is mostly unmaintained and adding even more modules will make this situation worse. The only thing that happens in this repo is I periodically go through the opened PRs and leave reviews/merge/close PRs from time to time and create new releases. I don't go through the issues so there's no real process for maintaining everything in here.

This is not in any way a reflection on the code here, in fact I would say this is probably a good module to have, but I need to come up with a solid policy to state what will be and won't be accepted in future PRs to this collection vs what should be in people's own collections. I'll bring it up in the next windows working group to see what some of the other Windows maintainers have to say about this topic.

plugins/modules/win_listen_ports_facts.py
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2022, DataDope (@datadope-io)
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+DOCUMENTATION = r'''
+---
+module: win_listen_ports_facts
```
jborean93, line 9: Need to specify when this module was added.

```suggestion
module: win_listen_ports_facts
version_added: '1.10.0'
```

plugins/modules/win_listen_ports_facts.py
```diff
@@ -0,0 +1,93 @@
...
+  other than Listen
+seealso:
+- module: community.general.listen_ports_facts
+author:
+- David Nieto (@david-ns)
+'''
+
+EXAMPLES = r'''
+- name: Recopilate ports facts
+  ansible.windows.win_listen_ports_facts:
```
jborean93, line 47: The FQCN is `community.windows`:

```suggestion
  community.windows.win_listen_ports_facts:
```

plugins/modules/win_listen_ports_facts.py
```diff
@@ -0,0 +1,93 @@
...
+
+- name: Retrieve only ports with Closing and Established states
+  ansible.windows.win_listen_ports_facts:
```
jborean93, line 50: Ditto:

```suggestion
  community.windows.win_listen_ports_facts:
```

plugins/modules/win_listen_ports_facts.py
```diff
@@ -0,0 +1,93 @@
...
+    tcp_filter:
+        - Closing
+        - Established
+
+- name: Get ports facts with only the year within the date field
+  ansible.windows.win_listen_ports_facts:
```
jborean93, line 56: Ditto:

```suggestion
  community.windows.win_listen_ports_facts:
```

tests/integration/targets/win_listen_ports_facts/aliases
```diff
@@ -0,0 +1 @@
+windows/adds/group1
```
jborean93, line 1: This isn't a valid group, please use

```suggestion
shippable/windows/group5
```

Are you also able to let me know where you got this value? I've seen a few PRs with this type of format and am wondering if our docs are out of date.

jborean93 (changes requested):

plugins/modules/win_listen_ports_facts.py
david-ns: I've added 'version_added' key

plugins/modules/win_listen_ports_facts.py
david-ns: Changed all the `ansible.windows.win_listen_ports_facts` to `community.windows.win_listen_ports_facts`

plugins/modules/win_listen_ports_facts.py
david-ns: Changed all the `ansible.windows.win_listen_ports_facts` to `community.windows.win_listen_ports_facts`

plugins/modules/win_listen_ports_facts.py
david-ns: Changed all the `ansible.windows.win_listen_ports_facts` to `community.windows.win_listen_ports_facts`

tests/integration/targets/win_listen_ports_facts/aliases
david-ns: Group updated. I got it from other modules, as I couldn't find in the docs where this value was from.

plugins/modules/win_listen_ports_facts.ps1
```diff
@@ -0,0 +1,81 @@
...
+
+# Structure of the response the script will return
+$ansibleFacts = @{
+    tcp_listen = @()
+    udp_listen = @()
+}
+
+# Build an index of the processes based on the PID
+$processes = @{}
+Get-Process -IncludeUserName | ForEach-Object {
```
jborean93, line 29: Looks like PowerShell 3.0 does not have the `-IncludeUserName` parameter which is why CI is failing. There are 3 things I can see you do to solve this problem

* Conditionally add this parameter and `user` return value

```powershell
$getProcCommand = Get-Command -Name Get-Process
$getProcParams = @{}
if ('IncludeUserName' -in $getProcCommand.Parameters.Keys) {
    $getProcParams.IncludeUserName = $true
}
Get-Process @getProcParams | ForEach-Object { ... }
```

You also need to update `Build-Listener` to handle `UserName` not being present or just update the docs saying `user` will be null on hosts with PowerShell 3.0.

* Mark the module as needing PowerShell 4.0 or newer in the docs and skip 2012 in the tests `skip/windows/2012` in the aliases file

The docs would need to updated to reflect this.

* Use PInvoke to get the username from the process yourself

This is quite involved but luckily we have some helper utils to do this for you

```csharp
#AnsibleRequires -CSharpUtil Ansible.AccessToken

Function Get-ProcessUserName {
    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSAvoidUsingEmptyCatchBlock", "",
        Justification = "Getting the username is optional so errors are ignored")]
    [CmdletBinding()]
    param (
        [Parameter(Mandatory)]
        [int]
        $Id
    )

    try {
        $procHandle = [Ansible.Basic.TokenUtil]::OpenProcess($Id, "QueryInformation", $false)
        $token = [Ansible.Basic.TokenUtil]::OpenProcessToken($procHandle, "Query")
        $user = [Ansible.Basic.TokenUtil]::GetTokenUser($token)
        $user.Translate([System.Security.Principal.NTAccount]).Value
    }
    catch {}  # Most likely doesn't have enough rights, just silently ignore
}
```

david-ns: Thanks for the information @jborean93. I've followed a different approach to fix the missing `-IncludeUserName` functionality: replacing the `Get-Process` cmdlet with `Get-CimInstance -ClassName Win32_Process` and using `GetOwner`, so no compromise is needed (no conditional `user` returning, no PowerShell 4 requierement or not a possible silent error).

david-ns:
@jborean93 it seems like the CI is failing (`ERROR! Error when getting collection version metadata for chocolatey.chocolatey:1.2.0 from default (https://galaxy.ansible.com/api/) (HTTP Code: 429, Message: Too Many Requests Code: Unknown)`). Excluding the four tests that failed because of that error, it seems like no other error is raised.

jborean93:
Yea that's an annoying problem I need to fix up. I've rerun the failed test and looks like 2012 still has other problems.

plugins/modules/win_listen_ports_facts.ps1
david-ns: It seems like Powershell 3 has no 'Bound' State for the TCP ports. I'll update the docs to specify it and the tests so 'Bound' is not one of the states used to test globally.

david-ns:
> Yea that's an annoying problem I need to fix up. I've rerun the failed test and looks like 2012 still has other problems.

It seems like Powershell 3 has no 'Bound' State for the TCP ports. I'll update the docs to specify it and the tests so 'Bound' is not one of the states used to test globally.

jborean93 (approved):

PR #391, opened 2022-05-03
add automated docs building and publishing

briantist:
##### SUMMARY

As discussed: https://github.com/ansible/community/issues/644#issuecomment-1116309488

~I will wait until we're sure it's working before adding a new PR deleting the RST files in `docs/` and the related workflow.~
Follow-up PR is up: https://github.com/ansible-collections/community.windows/pull/392

##### ISSUE TYPE

- Docs Pull Request

### DIFF
--- a/.github/workflows/docs-pr.yml
+++ b/.github/workflows/docs-pr.yml
@@ -1,0 +1,91 @@
+name: Collection Docs
+concurrency:
+  group: docs-pr-${{ github.head_ref }}
+  cancel-in-progress: true
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened, closed]
+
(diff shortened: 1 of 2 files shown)


briantist:
Following discussion in WWG, I've modified the PR workflow to remove the publishing step, which is the step that runs with `contents: write` permission.

What's left is the build job (read-only, only part that touches PR code), and the separate job for posting a comment on the PR, and doesn't have `contents` access at all.

This will let us still:
- verify that PRs are not causing docs problems
- post a helpful comment showing (in diff format) what's changed, and where to download the rendered docs for local review

---

The push/schedule workflow remains the same, with build + publish

PR #398, opened 2022-05-21
Add GUID mapping to fix AuditPol localization issue

### LINKED ISSUE #14: win_audit_policy_system does not work in non-English environment
_From @ssato on Mar 03, 2020 12:22_

##### SUMMARY

win_audit_policy_system does not work as expected in non-English environment
because it calls auditpol.exe and its output varies depends on the locale.

```
# sample task:
- name: set audit policies
  win_audit_policy_system:
    subcategory: "{{ item.name }}"
    audit_type: "{{ item.audit_type }}"
  loop: "{{ soe_audit_policies }}"
```

```
# var def.
soe_audit_policies:
  - name: Credential Validation
    audit_type: success, failure
```

```
# error message
failed: [xxx-win2016-vmware-soe] (item={u'audit_type': u'success, failure', u'name': u'Credential Validation'}) => {"ansible_loop_var": "item", "changed": false, "item": {"audit_type": "success, failure", "name": "Credential Validation"}, "msg": "Get-AnsibleParam: Argument subcategory needs to be one of IPsec ,IPsec ,IPsec ,IPsec ,/ ,SAM,DPAPI ,RPC ,Token Right Adjusted Events,MPSSVC ,Kerberos ,Kerberos  but was Credential Validation."}
```

The reason of this error looks because the line does not work at all as expected
in non English locale environment, that is, audtpol.exe outputs localized
non English results instead of English ones the powershell script expects.

```
$categories = ConvertFrom-Csv $(auditpol /list /category /r) \| Select-Object -expand Category*
```
[win_audit_policy_system.ps1#L27](https://github.com/ansible/ansible/blob/stable-2.9/lib/ansible/modules/windows/win_audit_policy_system.ps1#L27)]

```
# An example of outputs from autitpol.exe in English (US) locale environment 
> ConvertFrom-Csv $(auditpol /list /category /r)
Category/Subcategory GUID
-------------------- ----
DS                   {699xxxxx-xxxx-xxxx-xxxxxxxxxxxxxxxx}
                        {699xxxxx-xxxx-xxxx-xxxxxxxxxxxxxxxx}
                        {699xxxxx-xxxx-xxxx-xxxxxxxxxxxxxxxx}
                        ... (snip) ....
/                       {699xxxxx-xxxx-xxxx-xxxxxxxxxxxxxxxx}
                        ... (snip) ...
```

```
# An example of outputs from autitpol.exe in Japanese (JP) locale environment 
> ConvertFrom-Csv $(auditpol /list /category /r)
カテゴリ/サブカテゴリ GUID
--------------------- ----
DS アクセス                  {699xxxxx-xxxx-xxxx-xxxxxxxxxxxxxxxx}
アカウント ログオン     {699xxxxx-xxxx-xxxx-xxxxxxxxxxxxxxxx}
アカウント管理             {699xxxxx-xxxx-xxxx-xxxxxxxxxxxxxxxx}
オブジェクト アクセス {699xxxxx-xxxx-xxxx-xxxxxxxxxxxxxxxx}
システム                        {699xxxxx-xxxx-xxxx-xxxxxxxxxxxxxxxx}
ポリシーの変更             {699xxxxx-xxxx-xxxx-xxxxxxxxxxxxxxxx}
x`ログオン/ログオフ        {699xxxxx-xxxx-xxxx-xxxxxxxxxxxxxxxx}
特権の使用                    {699xxxxx-xxxx-xxxx-xxxxxxxxxxxxxxxx}
詳細追跡                        {699xxxxx-xxxx-xxxx-xxxxxxxxxxxxxxxx}
```

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

win_audit_policy_system

##### ANSIBLE VERSION

```
$ ansible --version
ansible 2.9.2
  config file = None
  configured module search path = [u'/home/vagrant/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /home/vagrant/.local/lib/python2.7/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 2.7.5 (default, Jun 20 2019, 20:27:34) [GCC 4.8.5 20150623 (Red Hat 4.8.5-36)]

$
```

##### CONFIGURATION

```paste below
```
(nothing was output.)

##### STEPS TO REPRODUCE

See the previous example.

##### EXPECTED RESULTS

It works as expected.

##### ACTUAL RESULTS

See the previous example.

_Copied from original issue: ansible/ansible#67962_

clcaldwell:
##### SUMMARY

Added a static GUID - English Name mapping for the outputs from auditpol.exe, to get around an issue caused by auditpol's outputs are localized. 

Fixes #14

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

win_audit_policy_system

##### ADDITIONAL INFORMATION
This is the same method used to fix the same localization problem is the AuditPolicyDSC DSC resource:
https://github.com/dsccommunity/AuditPolicyDsc/blob/03bd14a560be01a50222f1dfad739c914043b124/DSCResources/AuditPolicyResourceHelper/AuditPolicyResourceHelper.psm1#L249

### DIFF
--- a/plugins/modules/win_audit_policy_system.ps1
+++ b/plugins/modules/win_audit_policy_system.ps1
@@ -23,7 +23,8 @@ $categories_rc = run-command -command 'auditpol /list /category /r'
 $subcategories_rc = run-command -command 'auditpol /list /subcategory:* /r'
 
 If ($categories_rc.item('rc') -eq 0) {
-    $categories = ConvertFrom-Csv $categories_rc.item('stdout') | Select-Object -expand Category*
+    $categories_guid = ConvertFrom-Csv $categories_rc.item('stdout') | Select-Object -expand GUID
+    $categories_guid = $categories_guid.Trim('{', '}')
 }
 Else {
     Fail-Json -obj $results -message "Failed to retrive audit policy categories. Please make sure the auditpol command is functional on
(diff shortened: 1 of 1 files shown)


jborean93:
I'm not sure hardcoding in a list of categories and sub categories is the best idea here. IMO this module should accept 3 forms for the category and subcategory options:

* Raw GUID, allows someone to bypass all localization and specify the GUID as is
* The localized name, allows someone to enter a more user friendly and location specific setting
* If the above 2 fail then have the fallback mapping from the English name to the known GUID

I'm not 100% convinced of the 3rd one but I'm not against it. I can see how it could be convenient to have an Ansible play with the English label that works across any system but I do think we should still accept the localized variant and only have the English value as a final fallback.

As for the localization issue with `auditpol` it seems the better method would be to map all human readable labels of the `category` and `subcategory` to the GUID and call the auditpol commands with the GUID specified rather than the label. This keeps the commands relative simple as we don't need to quote the value and it avoids any localization problems.

To get a list of all categories and subcategories, including their localized names you can use the following code that utilities PInvoke to call the relevant Win32 function.

<details>
  <summary>Click to expand!</summary>

```powershell
Add-Type -TypeDefinition @'
using System;
using System.ComponentModel;
using System.Runtime.InteropServices;

namespace Community.Windows.WinAuditPolicySystem
{
    public class Native
    {
        [DllImport("Advapi32.dll")]
        private static extern void AuditFree(IntPtr Buffer);
            
        [DllImport("Advapi32.dll", EntryPoint = "AuditEnumerateCategories", SetLastError = true)]
        [return: MarshalAs(UnmanagedType.I1)]
        private static extern bool NativeAuditEnumerateCategories(
            out IntPtr ppAuditCategoriesArray,
            out int pdwCountReturned);

        public static Guid[] AuditEnumerateCategories()
        {
            IntPtr catArray = IntPtr.Zero;
            int catCount = 0;
            if (!NativeAuditEnumerateCategories(out catArray, out catCount))
            {
                throw new Win32Exception();
            }
            
            try
            {
                IntPtr guidPtr = catArray;
                Guid[] categories = new Guid[catCount];
                for (int i = 0; i < catCount; i++)
                {
                    categories[i] = (Guid)Marshal.PtrToStructure(guidPtr, typeof(Guid));
                    guidPtr = IntPtr.Add(guidPtr, Marshal.SizeOf(typeof(Guid)));
                }
                
                return categories;
            }
            finally
            {
                AuditFree(catArray);
            }
        }

        [DllImport("Advapi32.dll", EntryPoint = "AuditEnumerateSubCategories", SetLastError = true)]
        [return: MarshalAs(UnmanagedType.I1)]
        private static extern bool NativeAuditEnumerateSubCategories(
            ref Guid pAuditCategoryGuid,
            [MarshalAs(UnmanagedType.I1)] bool bRetrieveAllSubCategories,
            out IntPtr ppAuditCategoriesArray,
            out int pdwCountReturned);

        public static Guid[] AuditEnumerateSubCategories(Guid auditCategory)
        {
            IntPtr catArray = IntPtr.Zero;
            int catCount = 0;
            if (!NativeAuditEnumerateSubCategories(ref auditCategory, false, out catArray, out catCount))
            {
                throw new Win32Exception();
            }
            
            try
            {
                IntPtr guidPtr = catArray;
                Guid[] categories = new Guid[catCount];
                for (int i = 0; i < catCount; i++)
                {
                    categories[i] = (Guid)Marshal.PtrToStructure(guidPtr, typeof(Guid));
                    guidPtr = IntPtr.Add(guidPtr, Marshal.SizeOf(typeof(Guid)));
                }
                
                return categories;
            }
            finally
            {
                AuditFree(catArray);
            }
        }

        [DllImport("Advapi32.dll", CharSet = CharSet.Unicode, SetLastError = true)]
        [return: MarshalAs(UnmanagedType.I1)]
        private static extern bool AuditLookupCategoryNameW(
            ref Guid pAuditCategoryGuid,
            out IntPtr ppszCategoryName);
            
        public static string AuditLookupCategoryName(Guid category)
        {
            IntPtr namePtr = IntPtr.Zero;
            if (!AuditLookupCategoryNameW(ref category, out namePtr))
            {
                throw new Win32Exception();
            }
            
            try
            {
                return Marshal.PtrToStringUni(namePtr);
            }
            finally
            {
                AuditFree(namePtr);
            }
        }

        [DllImport("Advapi32.dll", CharSet = CharSet.Unicode, SetLastError = true)]
        [return: MarshalAs(UnmanagedType.I1)]
        private static extern bool AuditLookupSubCategoryNameW(
            ref Guid pAuditCategoryGuid,
            out IntPtr ppszCategoryName);
            
        public static string AuditLookupSubCategoryName(Guid category)
        {
            IntPtr namePtr = IntPtr.Zero;
            if (!AuditLookupSubCategoryNameW(ref category, out namePtr))
            {
                throw new Win32Exception();
            }
            
            try
            {
                return Marshal.PtrToStringUni(namePtr);
            }
            finally
            {
                AuditFree(namePtr);
            }
        }
    }
}
'@

[Community.Windows.WinAuditPolicySystem.Native]::AuditEnumerateCategories() | ForEach-Object {
    $category = $_
    $name = [Community.Windows.WinAuditPolicySystem.Native]::AuditLookupCategoryName($category)
    $subCategories = [Community.Windows.WinAuditPolicySystem.Native]::AuditEnumerateSubCategories($_) | ForEach-Object {
        $subName = [Community.Windows.WinAuditPolicySystem.Native]::AuditLookupSubCategoryName($_)
        [PSCustomObject]@{
            Name = $subName
            Id = $_
        }
    }
    
    [PSCustomObject]@{
        Name = $name
        Id = $category
        SubCategories = $subCategories
    }
}
```
</details>

We can also still continue to use `auditpol /list ... /v` to get the info from the executable and still continue to work with the localized names as they can be mapped to the known GUIDs and vice versa.

clcaldwell:
I don't disagree with this, but I am having a hard time finding out exactly how the Ansible project has been handling localizations; I'm not seeing a whole lot of examples or clear guidance on localization issues.

[win_power_plan](https://github.com/ansible-collections/community.windows/blob/main/plugins/modules/win_power_plan.ps1) looks like it handles localization in a similar way to what you are proposing (it has a separate GUID field).

[win_timezone](https://github.com/ansible-collections/community.windows/blob/main/plugins/modules/win_timezone.ps1) is using tzutil.exe, which doesn't support localization (neither does Get-TimeZone), but there **are** localizations being supplied by the OS for time zones, which we could presumably grab from somewhere, see:
image 

It seems Ansible doesn't have the best localization support in general (ex. module files all have hardcoded English error messages). I don't mind modifying this PR with your suggestion, but I'm not entirely sure if it aligns with the rest of the Ansible project.

jborean93:
> I don't disagree with this, but I am having a hard time finding out exactly how the Ansible project has been handling localizations; I'm not seeing a whole lot of examples or clear guidance on localization issues.

There really isn't a clear guidance as it's not something we've really set out to define before. That's why you see so many different things across the various modules.

To me the question becomes, what would you like to input if you were writing a playbook. I see 3 different options

* The English name, this is most likely what people would default to but if English wasn't your first language then it's not as self documenting
* Your locale name, this is good for people who aren't English speakers as they can read a playbook as a person and understand what it's trying to set
* The unique ID (GUID) of the setting, this is completely unambiguous and should work

The programmer in me would be doing `category: 7d1880db-6d82-4047-8bb4-a9957655dcb4 # Category Name` but I can understand the UX isn't great on this. The benefits here is that this definition would work on all systems, regardless of their locale. It also helps to ensure the code itself is using the unique ids internally in it's logic making it less susceptible to localisation problems.

From a non-English user supporting English only isn't a great user experience. It could be just as opaque as a GUID if you only had to use English values for these fields.

Ultimately I've probably strayed too far off what you are trying to do. Let's just focus on getting the internals all working by mapping the input values to a known set of GUIDs and have the internal logic use that for comparisons to fix the localisation problems. If we want to talk about making it more friendly and usable to define categories in different languages we can focus on that in a different PR.

jborean93:
One thing I have noticed is that the code is still attempting to use the category names with `auditpol` in a few places.

https://github.com/ansible-collections/community.windows/pull/398/files#diff-4967b1679882e40bdd42fbd2af57f3afafd797048252bf6d59d2168188f411d5R181-L94

This to me seems like it will continue to have problems with localisation and these commands should be changed so it invokes with the GUID as the identifier making it work regardless of the locale.

briantist:
I also think it would be best to accept GUID _or_ localized form. I am leaning against the hardcoded English fallback.

One thing to consider is that there is not only one "English" to fall back to. "American" (US) English is used by a minority of the world's English-speaking population, but may be appear to be weighed more heavily in tech, while others (Australian English, British English, Indian English, etc.) are separate locales. Spelling and choice of terms can vary between any of them.

I am not sure how much the terms change for this, but I ran @jborean93 's code to get the localized variants and for example I see `Authorization Policy Change` (for `0cce9231-69ae-11d9-bed3-505054503030`) and I wonder if when he runs it, it's `Authorisation Policy Change` (a quick google suggests it probably still uses a `Z` because I can't find any results).

---

Anyway, point being, while it may work for this use case, hardcoding a list of names can be problematic. I would urge use of options 1 & 2 in Jordan's suggestion and discard option 3. Examples should be updated to show both uses, and show at least one non-English language usage (and imo at least one with non-Latin characters as it stands out more what is going on).

guyomog78:
Hello, 
In this case, it's to workaroud Windows localization! Auditpol is localized for each language (it's an very old tool)
For ansible usage, I always use English even I'm French ;-) no accents

Dylan-Bs:
Sometimes, there is no choice, in my company, they don't want to work with a computer in english ahah!
@guyomog78 What workaround have you find for it ?

guyomog78:
"Unfortunately, I have no workaround. I have suggested the capability of using GUIDs because they are standard in all languages, but it seems to be unavailable at the moment... Still waiting and seeing!"

PR #400, opened 2022-05-31
New module: win_package_facts

david-ns:
##### SUMMARY
New module win_package_facts to gather information about the packages of the machine

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
- win_package_facts

##### ADDITIONAL INFORMATION
- Gather the information of the packages of the machine.
- Supports x32 and x64 systems.
- The module goal is to replicate the functionality of the linux module package_facts (ansible.builtin.package_facts), maintaining the format of the said module.

### DIFF
--- a/plugins/modules/win_package_facts.ps1
+++ b/plugins/modules/win_package_facts.ps1
@@ -1,0 +1,172 @@
+#!powershell
+
+# Copyright: (c) 2022, DataDope (@datadope-io)
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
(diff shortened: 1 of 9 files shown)


jborean93:
As mentioned in https://github.com/ansible-collections/community.windows/pull/406, this collection is no longer accepting new module and plugin contributions. We appreciate the efforts put into the PR here but the time it takes to maintain the existing collections already present is not sustainable and accepting even more plugins will only make it worse. The beauty of collections is that these can be done in your own collection that you maintain and release as per your own schedule. Any user can publish their work to https://galaxy.ansible.com/ and share their work with others.

PR #413, opened 2022-07-07
fix docs build errors

briantist:
##### SUMMARY

With #391 merged, we have workflows building the docs and looking for issues.
This PR fixes the (very few!) latent issues in the docs currently in the collection.

Without this, all of the docs builds are going to fail even without introducing new errors, so this will get us to a clean place and then hopefully it will stay that way!

The warnings being fixed can be seen when the docs build does its base build:
https://github.com/ansible-collections/community.windows/runs/7226037816?check_suite_focus=true#step:8:336

##### ISSUE TYPE

- Docs Pull Request

### DIFF
--- a/plugins/modules/win_scheduled_task.py
+++ b/plugins/modules/win_scheduled_task.py
@@ -346,7 +346,7 @@ options:
   execution_time_limit:
     description:
     - The amount of time allowed to complete the task.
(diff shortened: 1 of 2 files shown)


briantist:
The CI failures are due to Galaxy 429 errors :(

jborean93 (approved):

Issue #424, opened 2022-08-11
win_psmodule error: A parameter cannot be found that matches parameter name 'AcceptLicense' in 1.11.0


shadowink:
##### SUMMARY
In 1.11.0 win_psmodule is throwing the error:Problems installing PSSlack module: A parameter cannot be found that matches parameter name 'AcceptLicense'.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
win_psmodule

##### ANSIBLE VERSION

```paste below
2.11.6
```

##### COLLECTION VERSION

```paste below
1.11.0
```

##### CONFIGURATION

```paste below

```

##### OS / ENVIRONMENT

Ansible running on a Win10 20h2 VM via a packer build.

##### STEPS TO REPRODUCE

```yaml
- name: Add Powershell Slack module
  community.windows.win_psmodule:
    name: PSSlack
    state: present

```

##### EXPECTED RESULTS

{
  "changed": true,
  "nuget_changed": true,
  "output": "Module PSSlack installed",
  "repository_changed": false
}

##### ACTUAL RESULTS

```paste below
"msg": "Problems installing PSSlack module: A parameter cannot be found that matches parameter name 'AcceptLicense'."
```
I downgrading the collection to 1.10.0 and the error did not occur.

jborean93:
What's the version of `Get-Module -Name PowerShellGet -ListAvailable` on the host in question? There is a minimum version requirement https://docs.ansible.com/ansible/latest/collections/community/windows/win_psmodule_module.html#notes of `1.6.0` which should be new enough to include the `-AcceptLicense`  parameter.

briantist:
re: feature check: I suppose we could do something like `(Get-Command Install-Module).Parameters.ContainsKey('AcceptLicense')` to check. 

I suspect that the issue is that I perhaps incorrectly added the parameter to the internal function that installs the pre-requisites:
https://github.com/ansible-collections/community.windows/blob/main/plugins/modules/win_psmodule.ps1#L95

So probably on this fresh host, it's trying to get the required modules to their minimums and hitting the problem there. Since those modules will never need the license check, we can probably just remove this line instead of feature testing. 

Ideally we'd have a test for this pre-req functionality but I'm not sure how much of a pain that will be to implement in CI, given that all the other tests install the minimum versions as a setup step. The test could only successfully run against a CI host that didn't run any of the other PowerShellGet tests I guess.

jborean93:
[...]

PR #425, opened 2022-08-12
win_psmodule - fix bootstrapping of PowerShellGet

jborean93:
##### SUMMARY
The bootstrapping code needs to only add `-AcceptLicense` to the splat if the underlying `Install-Module` version supports it. This also ensures that the pre-reqs will be sourced from the specified repository.

Fixes: https://github.com/ansible-collections/community.windows/issues/424

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_psmodule

### DIFF
--- a/plugins/modules/win_psmodule.ps1
+++ b/plugins/modules/win_psmodule.ps1
@@ -66,7 +66,8 @@ Function Install-PrereqModule {
         [Switch]$TestInstallationOnly,
         [bool]$AllowClobber,
         [Bool]$CheckMode,
-        [bool]$AcceptLicense
+        [bool]$AcceptLicense,
+        [string]$Repository
     )
 
     # Those are minimum required versions of modules.
@@ -92,7 +93,6 @@ Function Install-PrereqModule {
                         MinimumVersion = $PrereqModules[$Name]
                         Force = $true
                         WhatIf = $CheckMode
-                        AcceptLicense = $AcceptLicense
                     }
                     $installCmd = Get-Command -Name Install-Module
                     if ($installCmd.Parameters.ContainsKey('SkipPublisherCheck')) {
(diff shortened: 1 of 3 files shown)


plugins/modules/win_psmodule.ps1
```diff
@@ -460,13 +466,13 @@ if ($repo_user -and $repo_pass ) {
 }
 
 if ( ($allow_clobber -or $allow_prerelease -or $skip_publisher_check -or
-        $required_version -or $minimum_version -or $maximum_version) ) {
+        $required_version -or $minimum_version -or $maximum_version -or $accept_license) ) {
     # Update the PowerShellGet and PackageManagement modules.
     # It's required to support AllowClobber, AllowPrerelease parameters.
-    Install-PrereqModule -AllowClobber $allow_clobber -CheckMode $check_mode -AcceptLicense $accept_license
+    Install-PrereqModule -AllowClobber $allow_clobber -CheckMode $check_mode -AcceptLicense $accept_license -Repository $repo
```
briantist, line 472: I was thinking about the repository issue yesterday and whether or not to add it, but I think it might be best not to explicitly specify it. If the repo you supply is private it probably doesn't have the pre-reqs, so this will fail in those cases. If we left it blank, it will search all repositories, and then find it in the public one unless PSGallery was removed.

For our tests, we should remove PSGallery to avoid pulling from the internet (and to avoid the conflict of it being available in both).

jborean93: I’m not sure, if someone specifies a repository then I would argue that it should be used for everything, even pre reqs. Even then the pre reqs are only installed when the newer arguments have been requested so I would be surprised if this actual code is run a lot in the wild.

Happy to talk about it further before a new release is made to ensure we cover all our bases.

briantist: My best guess is that the code is run a lot in the wild, but only once for a given host, so it will affect mostly new hosts being brought up. With the repo change, if the first use of `win_psmodule` on the new host installs a module from PSGallery, there will be no issue:

```
# first time running on a new host, pre-reqs will be updated
- community.windows.win_psmodule:
    name: psprivilege
```

but if the first use installs from a private repo, the pre-req install will fail.

```
# first time running on a new host, pre-reqs can't be found
- community.windows.win_psmodule:
    name: ACMEModule
    repo: PSACME
```

PR #426, opened 2022-08-15
Cache collection requirements

jborean93:
##### SUMMARY
Attempt to cache the collection requirements stage to avoid falling afoul of API rate limiting by only doing it once.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
community.windows


jborean93:
Yea this is not going well, will have to pick this up at a later time but seems like cleaning the workspace might be removing the previous artifacts for the job. I've spent too much time on this already and need to move on.

PR #447, opened 2022-10-03
Update win_scheduled_task.py

Udayendu:
##### SUMMARY
- Path required to search and disable the task that already exists.

##### ISSUE TYPE
- Docs Pull Request

+label: docsite_pr

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/win_scheduled_task.py
+++ b/plugins/modules/win_scheduled_task.py
@@ -504,6 +504,7 @@ EXAMPLES = r'''
 
 - name: Disable a task that already exists
   community.windows.win_scheduled_task:
+    path: \TaskChecker
     name: TaskToDisable
     enabled: no
 


jborean93:
Is this path actually needed, the example is for a test in the root path, it's just in your situation you have a task in a sub folder needing the `path` option.

Udayendu:
> Is this path actually needed, the example is for a test in the root path, it's just in your situation you have a task in a sub folder needing the `path` option.

@jborean93 In general we need to have an option for path and having things at root is always not a possibility. Based on the project needs, we should have flexibility to choose the path the way we need. Hence it would be nice to have the path as an option. So see if we can merge this PR.

PR #449, opened 2022-10-04
Try to cache collection dep installs

jborean93:
##### SUMMARY
Testing for caching collection deps during test to avoid the sporadic errors we see today.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
testing


jborean93:
cc @briantist, I finally cracked it and got it working. Happy for any tweaks/critiques you think would make it better.

briantist (approved):
I don't know AZP well but this looks great to me!

PR #450, opened 2022-10-07
Various improvements / bugfixes around disks management

gillg:
##### SUMMARY
This PR group several changes around disks management :
* There is a bug in win_disk_facts if you try to use win32_disk_drive on some virtual systems where the disk serial number could be a non null but empty and common value for all the disks. There is a concrete exemple on Google Cloud Platform where all the disks have a serial number `"                  "` 😅 I use the correlation between disk "Index" and "DiskNumber" and it seems cover all the possible cases.
* Support partition_id wich is a super reliable key to select a partition in `win_format`
* Return the partition `unique_id` (what can be reused in `win_format`) at the end of `win_partition`
* Allows to delete a partition without knowing its ID, or number or whatever, filtering by type.
The usual case is when you initialize a GPT disk, windows automaticaly creates a partition "MSR", but it's not necessarly useful in your case, the module `win_initialize_disk` doesn't returns it (and it not aware of it), and that could make the way you select your partition number a little bit tricky. So having a way to really cleanup the disk in an idempotent way seems good.
This thing could be a little bit dangerous (but less than randomly delefe the partition "0" at each run), so we only allows it if the filter returns 1 partition.

##### Other module proposition
I also created a module **win_disk_mount**, I could share it but I don't have time to write any tests, make a perfect doc, examples, etc, so if you are interested someone should contribute a little bit.
This is the minimal doc to understand it :
```yaml
module: win_disk_mount
short_description: Mount a partition to a letter or folder
description:
    - Mount a partition to a letter or folder
options:
    disk_number:
        description:
            - Disk number where the partition is (partition_number mandatory)
        required: false
        type: int
    partition_number:
        description:
            - Partition number to mount (disk_number mandatory)
        required: false
        type: int
    partition_id:
        description:
          -  Partition unique ID to select
        required: false
        type: str
    path:
        description:
          -  Partition unique path to select
        required: false
        type: str
    state:
        description:
          -  Add or remove the mount point
        required: false
        type: str
        default: 'present'
    force_assign_letter:
        description:
          -  Force to attach a letter in addition to the folder mount path.
        required: false
        type: bool
        default: 'no'
    mount_path:
        description:
          -  Path to mount the volume. If not defined, will use the next available drive letter.
        required: false
        type: str
```

Minimal examples :
```yaml
    - name: Mount partition in D drive
      win_disk_mount:
        disk_number: 1
        partition_number: 1
        mount_path: "D:"

    - name: Mount partition in a folder (junction)
      win_disk_mount:
        disk_number: 1
        partition_number: 1
        mount_path: C:/mydata

    - name: Mount partition in the nex available drive letter
       win_disk_mount:
        disk_number: 1
        partition_number: 1
```
Let me know !


jborean93:
@gillg do you need help trying to resolve the sanity and test problems?

gillg:
> @gillg do you need help trying to resolve the sanity and test problems?

Thank you ! But I think I'm good now, I just forgot the CI result on this PR.

gillg:
`UniqueId` seems doesn't exists on some powershell versions... 😭 
Any idea ?

jborean93:
Not sure sorry, I don't really know much about these modules or the cmdlets it calls.

PR #452, opened 2022-10-24
Fix CNAME creation in win_dns_record

### LINKED ISSUE #429: win_dns_record - CNAME in a subdomain is not created properly
##### SUMMARY

I have tried to create a CNAME record in a subdomain. Instead of adding the record in the existing subdomain, the module recreate an arborescence of subdomain in which the record is created.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

win_dns_record 

##### ANSIBLE VERSION

```paste below
ansible 2.10.8
  config file = /opt/ansible/ansible/ansible.cfg
  configured module search path = ['/opt/ansible/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.9.2 (default, Feb 28 2021, 17:03:44) [GCC 10.2.1 20210110]
```

##### COLLECTION VERSION

```paste below
# /usr/lib/python3/dist-packages/ansible_collections
Collection        Version
----------------- -------
community.windows 1.3.0

# /opt/ansible/.ansible/collections/ansible_collections
Collection        Version
----------------- -------
community.windows 1.7.0
```

##### CONFIGURATION

```paste below
DEFAULT_BECOME(/opt/ansible/ansible/ansible.cfg) = False
DEFAULT_BECOME_ASK_PASS(/opt/ansible/ansible/ansible.cfg) = False
DEFAULT_BECOME_METHOD(/opt/ansible/ansible/ansible.cfg) = sudo
DEFAULT_BECOME_USER(/opt/ansible/ansible/ansible.cfg) = root
DEFAULT_HOST_LIST(/opt/ansible/ansible/ansible.cfg) = ['/opt/ansible/ansible/inventory']
DEFAULT_REMOTE_USER(/opt/ansible/ansible/ansible.cfg) = ansible
HOST_KEY_CHECKING(/opt/ansible/ansible/ansible.cfg) = False
```

##### STEPS TO REPRODUCE

Based on the following example :

- Domain : example.com
- Subdomain : test.example.com
- Target host : myserver.example.com
- The alias I want to create : helloworld.test.example.com

Use 

```yaml
- name: Create the server alias
      name: "helloworld.test"
      type: "CNAME"
      value: "myserver.example.com"
      zone: "example.com"
```

##### EXPECTED RESULTS

The expected result is the creation of a CNAME record named "helloworld" in the subdomain "test" of the domain "example.com". This record should point to the host myserver.example.com.

##### ACTUAL RESULTS

In the domain "example.com", I get a new subdomain named "com", under which I get a new subdomain named "example", under which I get a new subdomain named "test", under which I find the record "helloworld".

##### PROPOSED SOLUTION
After taking a look at the module code, I think the behaviour could be fixed by adding a specific case for CNAME in this [try/catch](https://github.com/ansible-collections/community.windows/blob/main/plugins/modules/win_dns_record.ps1#L151).
I have found there is a powershell module dedicated to CNAME record.

The following powershell command is able to create te required CNAME : `Add-DnsServerResourceRecordCName -Name "helloworld.test" -HostNameAlias "myserver.example.com" -ZoneName "example.com"`

So we could add something like : 
```
elseif ($type -eq 'CNAME') {
    Add-DnsServerResourceRecordCName -Name $name -HostNameAlias $value -ZoneName $zone
}
```
I think the extra arguments are also compatible with this powershell module, but I am not sure.

Change to documentation : 

- When the "type" parameter has "CNAME" has a value, the "name" parameter could include subdomain(s) where the record should be created, according to the domain defined by the "zone" parameter".
- When the "type" parameter has "CNAME" has a value, the "value" parameter contains the target host.

JLE-ATS:
##### SUMMARY

As suggested in https://github.com/ansible-collections/community.windows/issues/429, I have added a case to handle CNAME creation when the cname is part of a subdomain.
This is my first contribution and I am not used to the contribution process, so I am sorry if this is not the right way to submit fixes.

Fixes #429

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

win_dns_record

##### ADDITIONAL INFORMATION

If this fix is accepted, the documentation should be updated with instructions for creating aliases in subdomains.
Unfortunately, I didn't find where I could update the documentation in this repository.
The following task could be included in the examples on the documentation page :

```paste below
- name: Create the alias helloworld.test.example.com pointing to myserver.example.com
      name: "helloworld.test"
      type: "CNAME"
      value: "myserver.example.com"
      zone: "example.com"
```

### DIFF
--- a/plugins/modules/win_dns_record.ps1
+++ b/plugins/modules/win_dns_record.ps1
@@ -32,6 +32,7 @@ $values = $module.Params.value
 $weight = $module.Params.weight
 $zone = $module.Params.zone
 $dns_computer_name = $module.Params.computer_name
+# If you add additional arguments, check that they are supported by all modules (Add-DnsServerResourceRecordCName and Add-DnsServerResourceRecord).
 $extra_args = @{}
 if ($null -ne $dns_computer_name) {
     $extra_args.ComputerName = $dns_computer_name
@@ -156,7 +157,7 @@ if ($null -ne $values -and $values.Count -gt 0) {
                 Add-DnsServerResourceRecord -TXT -Name $name -DescriptiveText $value -ZoneName $zone -TimeToLive $ttl @extra_args -WhatIf:$module.CheckMode
             }
             elseif ($type -eq 'CNAME') {
-                Add-DnsServerResourceRecordCName -Name $name -HostNameAlias $value -ZoneName $zone -TimeToLive $ttl -WhatIf:$module.CheckMode
+                Add-DnsServerResourceRecordCName -Name $name -HostNameAlias $value -ZoneName $zone -TimeToLive $ttl @extra_args -WhatIf:$module.CheckMode
             }
             else {
                 Add-DnsServerResourceRecord -Name $name -AllowUpdateAny -ZoneName $zone -TimeToLive $ttl @splat_args -WhatIf:$module.CheckMode @extra_args
--- a/plugins/modules/win_dns_record.ps1
+++ b/plugins/modules/win_dns_record.ps1
@@ -150,10 +150,13 @@ if ($null -ne $values -and $values.Count -gt 0) {
         }
         try {
             if ($type -eq 'SRV') {
-                Add-DnsServerResourceRecord -SRV -Name $name  -ZoneName $zone @srv_args @extra_args -WhatIf:$module.CheckMode
+                Add-DnsServerResourceRecord -SRV -Name $name -ZoneName $zone @srv_args @extra_args -WhatIf:$module.CheckMode
             }
             elseif ($type -eq 'TXT') {
                 Add-DnsServerResourceRecord -TXT -Name $name -DescriptiveText $value -ZoneName $zone -TimeToLive $ttl @extra_args -WhatIf:$module.CheckMode
+            }
+            elseif ($type -eq 'CNAME') {
+                Add-DnsServerResourceRecordCName -Name $name -HostNameAlias $value -ZoneName $zone -TimeToLive $ttl -WhatIf:$module.CheckMode
             }
             else {
                 Add-DnsServerResourceRecord -Name $name -AllowUpdateAny -ZoneName $zone -TimeToLive $ttl @splat_args -WhatIf:$module.CheckMode @extra_args
--- a/changelogs/fragments/452-win_dns_record-cname.yml
+++ b/changelogs/fragments/452-win_dns_record-cname.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - win_dns_record - Fix CNAME creation in subdomains - https://github.com/ansible-collections/community.windows/issues/429


jborean93:
Just closing and re-opening to clear out the CI problems.

plugins/modules/win_dns_record.ps1
```diff
@@ -150,11 +150,14 @@ if ($null -ne $values -and $values.Count -gt 0) {
...
         try {
             if ($type -eq 'SRV') {
-                Add-DnsServerResourceRecord -SRV -Name $name  -ZoneName $zone @srv_args @extra_args -WhatIf:$module.CheckMode
+                Add-DnsServerResourceRecord -SRV -Name $name -ZoneName $zone @srv_args @extra_args -WhatIf:$module.CheckMode
             }
             elseif ($type -eq 'TXT') {
                 Add-DnsServerResourceRecord -TXT -Name $name -DescriptiveText $value -ZoneName $zone -TimeToLive $ttl @extra_args -WhatIf:$module.CheckMode
             }
+            elseif ($type -eq 'CNAME') {
+                Add-DnsServerResourceRecordCName -Name $name -HostNameAlias $value -ZoneName $zone -TimeToLive $ttl -WhatIf:$module.CheckMode
```
jborean93, line 159: So the changes here seem ok I'm just not sure what to do about the extra arguments like `@extra_args`, `-AllowUpdateAny` and possibly `@splat_args` that are in use in the original call.

`extra_args` contains the `-ComputerName` parameter which `Add-DnsServerResourceRecordCName` does have so it should be safe to add that. I think `splat_args` can probably be ignored as  it contains the CName info that this new cmdlet implicitly knows. I have no idea what `-AllowUpdateAny` update any does but the fact that `Add-DnsServerResourceRecord` also specified it makes me think this new call should.

jborean93 (changes requested):
Are you able to add a changelog fragment as per https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment.

jborean93:
Ignore the sanity failure, it will be fixed with https://github.com/ansible-collections/community.windows/pull/454 and is unrelated to your changes here.

JLE-ATS:
Thanks for your feedback.

- I have added the "@extra_args" to the command. I have also added a comment near the variable declaration to prevent ulterior breaking changes.

- I have not added the "-AllowUpdateAny" for now. This parameter has the same behaviour for both powershell modules so we could add it, but it allows any authenticated users to modify the DNS record which seems as a security problem. Maybe we could add a parameter to the ansible module to allow enabling/disabling this flag. **What's your thoughts ?**

- I have not added the "@splat_args" based on your comment and the fact that I did not find what this argument does. Is it doing anything at all ? 

- I have added a fragment as requested.

plugins/modules/win_dns_record.ps1
JLE-ATS: Thanks for your feedback.

- I have added the "@extra_args" to the command. I have also added a comment near the variable declaration to prevent ulterior breaking changes.

- I have not added the "-AllowUpdateAny" for now. This parameter has the same behaviour for both powershell modules so we could add it, but it allows any authenticated users to modify the DNS record which seems as a security problem. Maybe we could add a parameter to the ansible module to allow enabling/disabling this flag. What's your thoughts ?

- I have not added the "@splat_args" based on your comment and the fact that I did not find what this argument does. Is it doing anything at all ? 

- I have added a fragment as requested.

jborean93 (approved):

PR #201, opened 2021-02-19
adding support for txt in win_dns_record

sgruber94:
##### SUMMARY

Adding support for TXT Records

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
win_dns_record

##### ADDITIONAL INFORMATION

```
- name: Creating a TXT record with descriptive Text
  community.windows.win_dns_record:
    name: "test"
    state: present
    type: "TXT"
    value: "justavalue"
    zone: "example.com"

```

PR #461, opened 2022-11-16
win_dns_record: tests: add tests-TXT.yml

kenyon:
Looks like this was forgotten in #201.


briantist (commented):
looks pretty good to me!
only thing I'm unsure about is changing the license identifier, what's the context on that change?

kenyon:
@briantist The very next line in each of those files says GPL v3.0+, so these lines conflict, see my commit message: https://github.com/ansible-collections/community.windows/pull/461/commits/4f7bc5e3cf96ae36e4b27ffbbff4afab04e56b48

The v3.0+ line was there before the SPDX-License-Identifier line, so I figure the line that was there first takes precedence.

jborean93:
I would just keep the existing line for now and separate it out into a different PR. Will have to run it by some more knowledgable people to figure out whether we can actually make that change and which one takes precedence here and I don't want to hold up the test fix itself.

Just an FYI you don't need changelog fragments for test fixes.

kenyon:
@jborean93 OK, removed the changelog fragment commit and the license commit. Rebased and fixed merge conflict.

jborean93:
Looks like you included a commit from another PR in this now as it edits the module and not just the tests.

kenyon:
@jborean93 commit 678c9b332aa385e81894920adf19a9eacd527c04 was there before and is intentional. Updated PR title.

kenyon:
Just like in #464 where I removed unneeded special-case code for CNAME records, multiple contributors are not understanding how the code has cleverly handled the need for different powershell command line options.

kenyon:
Anything else needed to get this merged?

jborean93 (approved):

jborean93:
@kenyon apologies for the delay here. Thank you very much for working on these modules and also improving the tests, it is really appreciated. Unfortunately this one conflicts with another of your PRs I just merged. If you can rebase this I'm happy to merge it when it's ready.

kenyon:
@jborean93 thanks! Rebased, fixed the expected merge conflict, tests passed! 🚀

PR #464, opened 2022-11-20
win_dns_record: CNAME: remove unneeded special case

kenyon:
The special case for CNAME records added in #452 is not needed.


JLE-ATS:
Hello !
I am the person behind #452 
I have tested your solution by replacing the variable win_dns_record_zone with the value "example.com" and it works great for me too.
Thanks for your input.

kenyon:
Anything else needed to get this merged?

jborean93:
Sorry @kenyon for the delay, just getting back to your PRs. Thanks @JLE-ATS for confirming this fixes your problem and the original changes in #452 are no longer needed, that was going to be my next question :)

Issue #471, opened 2022-12-07
win_firewall_rule fails when program is set to any


PfurtschellerP:
##### SUMMARY

When using the `win_firewall_rule module` and having set program to `any` the module fails with the following message:

```
The full traceback is:
Value does not fall within the expected range.
At line:268 char:37
+ ...                                 $existingRule.$prop = $new_rule.$prop
+                                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : OperationStopped: (:) [], ArgumentException
    + FullyQualifiedErrorId : System.ArgumentException
```

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

win_firewall_rule

##### ANSIBLE VERSION

```paste below
ansible [core 2.13.6]
  config file = None
  configured module search path = ['/home/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/user/.local/lib/python3.9/site-packages/ansible
  ansible collection location = /home/user/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/user/.local/bin/ansible
  python version = 3.9.2 (default, Feb 28 2021, 17:03:44) [GCC 10.2.1 20210110]
  jinja version = 3.1.2
  libyaml = True
```

##### COLLECTION VERSION

```paste below
Collection        Version
----------------- -------
community.windows 1.11.1
```

##### CONFIGURATION

```paste below

```

##### OS / ENVIRONMENT

Ansible execution from Debian 11 and AWX/RHEL8.6.
Target system is Windows Server 2019 Standard

##### STEPS TO REPRODUCE

- Try to set the "Program/ApplicationName" property on an existing firewall rule to "any"

```yaml
---
- name: Test
  hosts: targethost
  vars:
    fw_rules:
      - Name: Remote-Administration-RPC-EPMAP
        DisplayName: Remote Administration (RPC-EPMAP)
        DisplayGroup: Remote Administration
        Profile: Any
        Enabled: 'True'
        Action: Allow
        Program: Any # "%SystemRoot%\\system32\\svchost.exe"
        LocalAddress: Any
        RemoteAddress: Any
        Protocol: TCP
        LocalPort: RPC-EPMap
        RemotePort: Any
        Service: RPCSS
  tasks:
      - name: Apply firewall rules
        community.windows.win_test_firewall_rule:
          name: '{{ rule.DisplayName }}'
[...]

PR #472, opened 2022-12-07
Fix program can't be set to any

PfurtschellerP:
If ApplicationName of an existing rule should be set to any the parameter has to be set to $null which is not allowed. In these cases Set-NetFirewallRule can be used to set the parameter Program to "any" which translates to setting the ApplicationName of the rule to $null.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

win_firewall_rule

##### ADDITIONAL INFORMATION

https://github.com/ansible-collections/community.windows/issues/471

```paste below

```

### DIFF
--- a/plugins/modules/win_firewall_rule.ps1
+++ b/plugins/modules/win_firewall_rule.ps1
@@ -166,7 +166,11 @@ try {
     if ($null -ne $enabled) { $new_rule.Enabled = $enabled } else { $new_rule.Enabled = $true }
     if ($null -ne $description) { $new_rule.Description = $description }
     if ($null -ne $group) { $new_rule.Grouping = $group }
-    if ($null -ne $program -and $program -ne "any") { $new_rule.ApplicationName = [System.Environment]::ExpandEnvironmentVariables($program) }
+    if ($null -ne $program -and $program -ne "any") { 
+        $new_rule.ApplicationName = [System.Environment]::ExpandEnvironmentVariables($program) 
+    } elseif ($program -eq "any") {
+        $new_rule.ApplicationName = $program
+    }
     if ($null -ne $service -and $service -ne "any") { $new_rule.ServiceName = $service }
     if ($null -ne $protocol -and $protocol -ne "any") { $new_rule.Protocol = ConvertTo-ProtocolType -protocol $protocol }
     if ($null -ne $localport -and $localport -ne "any") { $new_rule.LocalPorts = $localport }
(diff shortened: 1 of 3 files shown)


plugins/modules/win_firewall_rule.ps1
```diff
@@ -166,7 +166,11 @@ try {
     if ($null -ne $enabled) { $new_rule.Enabled = $enabled } else { $new_rule.Enabled = $true }
     if ($null -ne $description) { $new_rule.Description = $description }
     if ($null -ne $group) { $new_rule.Grouping = $group }
-    if ($null -ne $program -and $program -ne "any") { $new_rule.ApplicationName = [System.Environment]::ExpandEnvironmentVariables($program) }
+    if ($null -ne $program -and $program -ne "any") { 
+        $new_rule.ApplicationName = [System.Environment]::ExpandEnvironmentVariables($program) 
+    } elseif ($program -eq "any") {
+        $new_rule.ApplicationName = $program
```
jborean93, line 172: This means that when someone sets `program: any` and a new rule is created the rule will have the program set `any` and not all like it does today. Where you trying to fix something by changing this logic?

jborean93 (changes requested):
I'm reluctant in bringin in `Set-NetFirewallRule`, especially since the powershell cmdlets don't have a way of getting a firewall rule by ID. This means if there are multiple rules with the same name they will all have the `Program` set rather than the one we are working on. In saying that I tried out the code and found that yes it fails to unset the ApplicationName.

Digging a bit deeper it seems like this is a fundamental problem with the COM binder in PowerShell and how it treats null values. Luckily there is a workaround, you can add the following to the top of the module

```powershell
#AnsibleRequires -PowerShell Ansible.ModuleUtils.AddType
```

Then after https://github.com/ansible-collections/community.windows/blob/f9157edc9ea202f4dd558153cf652a330cf72fcf/plugins/modules/win_firewall_rule.ps1#L131 add the following

```powershell
$_remote_tmp = Get-AnsibleParam $params "_ansible_remote_tmp" -type "path" -default $env:TMP
Add-CSharpType -TempPath $_remote_tmp -References @'
using System;
using System.Runtime.InteropServices;

namespace Community.Windows.WinFirewallRule
{
    [ComImport]
    [InterfaceType(ComInterfaceType.InterfaceIsIDispatch)]
    [Guid("AF230D27-BABA-4E42-ACED-F524F22CFCE2")]
    public interface INetFwRule
    {
        [DispId(3)]
        string ApplicationName { get; set; }
    }

    public static class NetFwRule
    {
        public static void PutApplicationName(object rule, string value)
        {
            ((INetFwRule)rule).ApplicationName = string.IsNullOrEmpty(value) ? null : value;
        }
    }
}
'@
```

Now when the ApplicationName needs to be set it can do it with

```powershell
ElseIf(($prop -eq 'ApplicationName') {
    $value = if ($new_rule.$prop -eq "any") { $null } else { $new_rule.$prop }
    [Community.Windows.WinFirewallRule.NetFwRule]::PutApplicationName($existingRule, $value)
}
```

It would also be great if you could add a test for this scenario to https://github.com/ansible-collections/community.windows/tree/f9157edc9ea202f4dd558153cf652a330cf72fcf/tests/integration/targets/win_firewall_rule.

plugins/modules/win_firewall_rule.ps1
PfurtschellerP: Yes, `Set-NetFirewallRule` accepts `any` for the program parameter and nulls the property. I won't be neccessary anymore when applying your solution.

PfurtschellerP:
@jborean93 is this now correct?

jborean93 (approved):
Yes thank you very much for implementing the new changes and writing the test for this specific scenario.

PR #480, opened 2022-12-22
DNS record - add possibility to enable aging

PfurtschellerP:
##### SUMMARY

The current `win_dns_record` module only creates static dns records.
The proposed change would allow for aging to be activated on the record (if the targeted zone supports it).

This unfortunately can't be set using `Set-DnsServerResourceRecord` (and `Set-DnsServerResourceRecordAging` only works one-way). That is why if a mismatch between what is requested and what is present is detected, the current record is removed and a new one is created.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

win_dns_record


PfurtschellerP:
Comment to remove stale label.

plugins/modules/win_dns_record.py
```diff
@@ -52,6 +52,12 @@
       dynamically "round up" other values to that minimum.
     default: 3600
     type: int
+  aging:
+    description:
+    - Should aging be activated for the record.
+    - If set to C(false), the record will be static.
+    default: false
+    type: bool
```
jborean93, line 60: Need to add `version_added: 1.13.0` to document when this option was added.

jborean93:
Thanks for the PR, would it be possible to add this scenario as a test to https://github.com/ansible-collections/community.windows/tree/main/tests/integration/targets/win_dns_record? Are you also able to add a changelog fragment to the PR https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment to help document the changes when generated the next release changelog.

From a technical perspective the change makes sense but I'm no expert in DNS records and these cmdlets. @kenyon if you get a chance it would be great to see your perspective on this change.

kenyon (changes requested):
I haven't had need to use this aging feature, but it looks useful. Looks good, but definitely would like to see tests.

kenyon (changes requested):
would be better if you could put the formatting changes in a separate commit

PfurtschellerP:
Okay. Added tests for A records. Are these sufficient enough? And should I add them to the other record types as well?

jborean93:
Sorry for the delay, just added some minor doc tweaks but the rest looks good.

PR #481, opened 2023-01-04
Append custom LDAP attributes to the module Result object

pierre-claranet:
##### SUMMARY
Append attributes present in ``attributes`` option, with their value, to the module output.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
win_domain_user

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/win_domain_user-append_custom_attributes_to_output.yml
+++ b/changelogs/fragments/win_domain_user-append_custom_attributes_to_output.yml
@@ -1,0 +1,2 @@
+minor_changes:
+- win_domain_user - Append attributes present in ``attributes`` option, with their value, to the module output.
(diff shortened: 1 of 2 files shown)


jborean93:
Sorry for the delay in reviewing the module. Unfortunately this module is ultimately going to be deprecated and removed at a future point in time in favour of [microsoft.ad.user module](https://ansible-collections.github.io/microsoft.ad/branch/main/collections/microsoft/ad/user_module.html#ansible-collections-microsoft-ad-user-module). While the collection is not live yet it is designed to be a supported collection for all things Active Directory.

In saying that, one of the features that the new module is moving away from is returning all the results about a user object back to the user. This is done to make the modules simpler in operation, especially around check mode, and the recommended approach is to use the [microsoft.ad.object_info module](https://ansible-collections.github.io/microsoft.ad/branch/main/collections/microsoft/ad/object_info_module.html#ansible-collections-microsoft-ad-object-info-module) to return the attributes you desire. There's even a copy of `object_info` in this module which can be used today.

pierre-claranet:
Ok, sorry, I wasn't aware of this...

Thanks for the info about the new collection and its object_info module, we'll switch to it.

PR #482, opened 2023-01-09
win_xml: Prevent issue when no childnode is defined

RyanBijkerk:
##### SUMMARY
It is not possible to add a new element to an empty element. To prevent this I have added logic to add an element when the node does not contain any childnodes. Please take a look at the additional information for more details and examples.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_xml

##### ADDITIONAL INFORMATION
My goal was to add a simple httpRedirect configuration in an existing web.config of an IIS server. In my example, I only had an empty configuration as by base XML.

Playbook:
```YAML
- name: Create httpRedirect
  win_xml:
    path: C:\inetpub\wwwroot\web.config
    xpath: "/configuration/system.webServer"
    fragment: "<httpRedirect />"
```
Base XML:
```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <system.webServer>
  </system.webServer>
</configuration>
```

In this example, no changes are made. When adjusting the XML by adding a text element:

Base XML:
```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <system.webServer>
    example
  </system.webServer>
</configuration>
```

Result XML:
```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <system.webServer>
    example
  <httpRedirect /></system.webServer>
</configuration>
```

Behavior when using the pull request adjustment:

Base XML:
```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <system.webServer>
  </system.webServer>
</configuration>
```

Result:
```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <system.webServer>
    <httpRedirect />
  </system.webServer>
</configuration>
```

Tested this in two separate environments on both Windows Server 2019 and Windows Server 2022 with the ansible versions: 
2.10.17 & 2.13.7

### DIFF
--- a/plugins/modules/win_xml.ps1
+++ b/plugins/modules/win_xml.ps1
@@ -191,7 +191,15 @@ if ($type -eq "element") {
             if ($node.get_NodeType() -eq "Document") {
                 $node = $node.get_DocumentElement()
             }
-            $elements = $node.get_ChildNodes()
+            
+            if ($node.ChildNodes.Count -eq 0) {
+                $elements = @()
+                $elements += $node
+            }
+            else {
+                $elements = $node.get_ChildNodes()
+            }
+
             [bool]$present = $false
             [bool]$changed = $false
             if ($elements.get_Count()) {
--- a/plugins/modules/win_xml.ps1
+++ b/plugins/modules/win_xml.ps1
@@ -191,7 +191,7 @@ if ($type -eq "element") {
             if ($node.get_NodeType() -eq "Document") {
                 $node = $node.get_DocumentElement()
             }
-            
+
             if ($node.ChildNodes.Count -eq 0) {
                 $elements = @()
                 $elements += $node
--- a/plugins/modules/win_xml.ps1
+++ b/plugins/modules/win_xml.ps1
@@ -193,8 +193,7 @@ if ($type -eq "element") {
             }
 
             if ($node.ChildNodes.Count -eq 0) {
-                $elements = @()
-                $elements += $node
+                $elements = @($node)
             }
             else {
                 $elements = $node.get_ChildNodes()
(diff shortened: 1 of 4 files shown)


RyanBijkerk:
Can someone please explain why these Sanity tests fail?

plugins/modules/win_xml.ps1
```diff
@@ -191,7 +191,15 @@ if ($type -eq "element") {
             if ($node.get_NodeType() -eq "Document") {
                 $node = $node.get_DocumentElement()
             }
-            $elements = $node.get_ChildNodes()
+
+            if ($node.ChildNodes.Count -eq 0) {
+                $elements = @()
+                $elements += $node
```
jborean93, line 197: Can you just do `$elements = @($node)`?

jborean93 (changes requested):
Thanks for the PR, it would be great if you could add a changelog fragment to document this bugfix https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment.

It would also be great if you could add this scenario to the existing win_xml tests at https://github.com/ansible-collections/community.windows/tree/main/tests/integration/targets/win_xml. Having a test ensures that no future changes will break this particular scenario inadvertently which is key for a module like this.

plugins/modules/win_xml.ps1
RyanBijkerk: Correct! Updated the code.

RyanBijkerk:
Added both test and change log fragment.

jborean93 (approved):
Thanks!

PR #483, opened 2023-01-10
Add multi-domains support

pierre-claranet:
##### SUMMARY
Added multi-domains support:
* Added a module parameter `multi_domains`, when set to `yes`, the module will look for the user account in the whole Active Directory forest
* Added the ability to add/remove the user account to/from groups belonging to differents domains

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
win_domain_user


jborean93:
As mentioned in https://github.com/ansible-collections/community.windows/pull/481#issuecomment-1399707946, this module will be deprecated sometime soon and we are trying to avoid adding new features to it.

This might be a feature that could be implemented in the [microsoft.ad collection](https://github.com/ansible-collections/microsoft.ad) but it would have to be generic enough to work for more than just the user module.

pierre-claranet:
Hello @jborean93, I see your point. The "multi domain" feature main logic is generic (query an AD global catalog DC to get the target object). 
This PR was containing more changes because in case of the user module we are also managing groups, it is a specific and more complex case.
I'll probably try again by openning a new PR on the microsoft.ad.user module, and maybe other if I have the time.

PR #504, opened 2023-03-29
`win_dhcp_lease` add ipv6 functionality

marcb4:
##### SUMMARY
Currently it is not possible to use the module `win_dhcp_lease`  for IPv6 addresses.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
`win_dhcp_lease`

##### ADDITIONAL INFORMATION
There are different ways on how to achieve the IPv6 functionality. We could also instead of boolean for  `v6` use something else like `version` and demand strings like `ipv4` or `ipv6`.
Or instead of using an if statement for every Powershell cmdlet extract everything to a separate ipv6 module.
Use whatever you may like from my commit.
Note: this is currently fully untested as I don't have the resources to test this against a development dhcp server.


plugins/modules/win_dhcp_lease.py
```diff
@@ -86,6 +86,11 @@
       - Specifies the description for reservation being created.
       - Only applicable to l(type=reservation).
     type: str
+  v6:
+    description:
+      - Specifies if IPv6 should be used.
+    type: bool
+    default: false
```
jborean93, line 93: Needs `version_added: 1.13.0` so people know when this new option was added.

jborean93 (changes requested):
The changes look ok (with 1 suggestion) to me but there are still 2 things needed before this PR is approved and merged.

The first is to add some integration tests for IPv6, we currently have some tests at https://github.com/ansible-collections/community.windows/tree/main/tests/integration/targets/win_dhcp_lease/tasks, it would be great to add some for IPv6 to test at least some of the functionality here.

The second is to add a changelog fragment as documented here https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment. The header would be `minor_changes` and typically we would have it in the format

```yaml
minor_changes:
- win_dhcp_lease - short description of new feature here.
```

marcb4:
I can't figure out why the command `Add-DhcpServerv6Scope -Name TestNetworkv6 -Prefix 2001:4898:7020:1020::` fails in CI. 
Does the CI servers support IPv6? Did some research and apparently after installing the DHCP role, adding of a DHCPv6 scope seems possible.

marcb4:
I will completely rewrite my changes and open another pull request when it's fully working as there are way too many commits for now.

PR #520, opened 2023-06-09
Update win_xml.ps1

bct-timo-crabbe:
##### SUMMARY
Fixed bug: module would not add new child nodes to empty parent nodes

Moved code that will modify the xml node outside the check to find an existing node

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
module: win_xml

### DIFF
--- a/changelogs/fragments/520-fixed-empty-element-bug.yaml
+++ b/changelogs/fragments/520-fixed-empty-element-bug.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+  - win_xml - Fix issue where it was not possible to add a new element in empty element.
--- a/changelogs/fragments/520-fixed-empty-element-bug.yaml
+++ b/changelogs/fragments/520-fixed-empty-element-bug.yaml
@@ -1,2 +1,2 @@
-bugfixes:
-  - win_xml - Fix issue where it was not possible to add a new element in empty element.
+minor_changes:
+  - win_xml - Revert fix from pull request 482 and move element presence check outside of element matching code.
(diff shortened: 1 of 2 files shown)


jborean93 (changes requested):
The changes seem ok to me at a glance but the PR needs two things:

* A changelog fragment as documented by https://docs.ansible.com/ansible/latest/community/development_process.html#creating-changelog-fragments
* An integration tests added to https://github.com/ansible-collections/community.windows/blob/main/tests/integration/targets/win_xml/tasks/main.yml

This is a complex module so having an integration test for this scenario is key to ensure the issue does not regress in the future.

bct-timo-crabbe:
I added the changelog fragment as you requested.
There already are integration tests for this bug in the file you mention, lines 103-127.

jborean93:
> There already are integration tests for this bug in the file you mention, lines 103-127.

If there are already tests for this then how are they not failing? Sounds like the tests aren't covered the bug and more tests are needed.

bct-timo-crabbe:
It looks like this bug was already fixed by pull request #482. The way this bug is fixed in that pull request is applied in a very roundabout way, I would prefer to apply my solution.
What would be the best way to go forward? Should I update the changelog fragment to minor_changes and describe the change I have done?

jborean93:
I guess the question is for me, what's the benefits of updating the code if it's already fixed. Are there further fixes with your code here, is there some problems with the other PR.

bct-timo-crabbe:
The fix in the other PR does not fix the root cause of the problem, it only forces the code to execute properly by adding extra logic that isn't necessary. There would be some performance benefits, but these would be negligible.

jborean93:
> The fix in the other PR does not fix the root cause of the problem, it only forces the code to execute properly by adding extra logic that isn't necessary.

From my perspective it fixed the issue, you even stated there is already a test for it that was added by the PR and it passes. My concern is that while your fix might be better the other fix might also solve other problems. Unless there is a clear reason why reverting the previous PR will help I'm not sure I can accept this PR sorry.

Issue #265, opened 2021-06-23
win_partition and bad behaviour with windows failover cluster


lenhof:
Hi,

It seems that when enabling failover cluster on a windows server trigger a race which broke win_format on the second run

Services to enable for reproduction :
  - "Failover-clustering"
  - "RSAT-Clustering-Mgmt"
  - "RSAT-Clustering-PowerShell"
  - "RSAT-Clustering-CmdInterface"

Here is the log trace :

```
Using module file /root/.ansible/collections/ansible_collections/community/windows/plugins/modules/win_partition.ps1
Pipelining is enabled.
EXEC (via pipeline wrapper)
The full traceback is:
Method invocation failed because [System.Object[]] does not contain a method named 'op_Subtraction'.
At line:233 char:17
+ ...         if ($ansible_partition_size -ne $ansible_partition.Size -and  ...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (op_Subtraction:String) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : MethodNotFound
ScriptStackTrace:
at <ScriptBlock>, <No file>: line 233
failed: [mssccstest1801] (item={'disk_number': 1, 'part_type': 'mbr', 'drive_letter': 'E', 'label': 'Data', 'file_system': 'NTFS', 'partition_size': -1, 'full_format': 'no'}) => {
    "ansible_loop_var": "disk_loop",
    "changed": false,
    "disk_loop": {
        "disk_number": 1,
        "drive_letter": "E",
        "file_system": "NTFS",
        "full_format": "no",
        "label": "Data",
        "part_type": "mbr",
        "partition_size": -1
    },
    "msg": "Unhandled exception while executing module: Method invocation failed because [System.Object[]] does not contain a method named 'op_Subtraction'."
}
```

jborean93:
Fixed by https://github.com/ansible-collections/community.windows/pull/543

PR #543, opened 2023-11-30
Fix issue #265. Support for Windows Failover Cluster

xcom1710:
##### SUMMARY
Fix issue #265. Support for Windows Failover Cluster

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_format.ps1
win_partition.ps1

##### ADDITIONAL INFORMATION
On systems with an active failover cluster Get-Partitions and Get-Volumes show entries from all nodes in that cluster.
To solve the problem the code filters out all partitions and volumes without a disk number.
AFAIK only a local disk will ever have a disk number.
Tested with Windows Server 2022.

### DIFF
--- a/changelogs/fragments/543-support_for_failover_cluster.yaml
+++ b/changelogs/fragments/543-support_for_failover_cluster.yaml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - Add support for windows failover cluster disks
--- a/changelogs/fragments/543-support_for_failover_cluster.yaml
+++ b/changelogs/fragments/543-support_for_failover_cluster.yaml
@@ -1,3 +1,3 @@
 ---
 bugfixes:
-  - Add support for windows failover cluster disks
+  - win_format, win_partition - Add support for Windows failover cluster disks
--- a/plugins/modules/win_format.ps1
+++ b/plugins/modules/win_format.ps1
@@ -74,7 +74,8 @@ function Get-AnsibleVolume {
 
     if ($null -ne $DriveLetter) {
         try {
-            $volume = Get-Volume -DriveLetter $DriveLetter
+            $partition = Get-Partition -DriveLetter $DriveLetter | Where-Object { $null -ne $_.DiskNumber }
+            $volume = Get-Volume -Partition $partition
         }
         catch {
             $module.FailJson("There was an error retrieving the volume using drive_letter $($DriveLetter): $($_.Exception.Message)", $_)
--- a/plugins/modules/win_format.ps1
+++ b/plugins/modules/win_format.ps1
@@ -74,6 +74,11 @@ function Get-AnsibleVolume {
 
     if ($null -ne $DriveLetter) {
         try {
+            # This needs to be a two step process so that we can support Windows failover cluster disks.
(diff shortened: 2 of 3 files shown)


changelogs/fragments/543-support_for_failover_cluster.yaml
```diff
@@ -0,0 +1,3 @@
+---
+bugfixes:
+  - Add support for windows failover cluster disks
```
jborean93, line 3: Could you reference the modules that this fixes please, something like `- win_format, win_partition - Add support for Windows failover cluster disks`.

plugins/modules/win_format.ps1
```diff
@@ -74,7 +74,8 @@ function Get-AnsibleVolume {
 
     if ($null -ne $DriveLetter) {
         try {
-            $volume = Get-Volume -DriveLetter $DriveLetter
+            $partition = Get-Partition -DriveLetter $DriveLetter | Where-Object { $null -ne $_.DiskNumber }
+            $volume = Get-Volume -Partition $partition
```
jborean93, line 78: Can you explain how this fixes the issue? When would there be a partition without a DiskNumber and why should we be ignoring them?

plugins/modules/win_partition.ps1
```diff
@@ -100,7 +100,7 @@ if ($null -ne $disk_number -and $null -ne $partition_number) {
 # Check if drive_letter is either auto-assigned or a character from A-Z
 elseif ($drive_letter -and $drive_letter -ne "auto" -and -not ($disk_number -and $partition_number)) {
     if ($drive_letter -match "^[a-zA-Z]$") {
-        $ansible_partition = Get-Partition -DriveLetter $drive_letter -ErrorAction SilentlyContinue
+        $ansible_partition = Get-Partition -DriveLetter $drive_letter -ErrorAction SilentlyContinue | Where-Object { $null -ne $_.DiskNumber }
```
jborean93, line 103: Same thing here, can you explain how this fixes the issue? When would there be a partition without a DiskNumber and why should we be ignoring them?

plugins/modules/win_format.ps1
xcom1710: In scenarios where the failover cluster role is installed on a windows server, Get-Volume and Get-Partition show every disk from every node participating in that cluster. Say we have a clustered drive M. Get-Volume would show you n many entries for drive M. The key differentiator between all those entries being the disk number. Only local disk will ever have a disk number. So with that in mind we can ignore all other disk that do not posses a disk number. The new code makes sure that we will always use the local disk as all relevant calls in the modules only work with local disk.
Hope this helps.

jborean93: Thanks for the explanation, could you possible include a comment that briefly explains this so at least we know not to remove those lines in the future? Same for the other file.

benjamin-rousseau-shift: @xcom1710 could you add the comment so that this PR can be merged T_T 
Also there is the same issue with win_disk_facts

xcom1710: done

jborean93 (approved):

PR #546, opened 2023-12-20
win_regmerge: Add support for specifying registry file contents directly as 'content'

d-a-n-d-u:
##### SUMMARY
Add support for specifying registry file contens directly using a new 'content' parameter instead of the existing 'path' parameter. If 'content' is not specified the behavior is same as before, registry file is merged using 'path'.

There's alos a bug in the documentation which lists 'compare_key' as an option/parameter when in fact the parameter is 'compare_to', fix included in this PR.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
win_regmerge

##### ADDITIONAL INFORMATION
Tested on: Windows Server 2022.

### DIFF
--- a/tests/integration/targets/win_regmerge/tasks/main.yml
+++ b/tests/integration/targets/win_regmerge/tasks/main.yml
@@ -140,7 +140,7 @@
 
 # test 1 -  basic test of changed behaviour
 # merge in REG_SZ
-- name: test 1 merge in a setting
+- name: test 1 merge in a setting using 'content'
   win_regmerge:
     content: "{{ lookup('file', 'settings1c.reg') }}"
   register: merge11_result
@@ -150,7 +150,7 @@
       - "merge11_result.changed == true"
 
 # re run the merge
-- name: test 1 merge in the setting again
+- name: test 1 merge in the setting again using 'content'
   win_regmerge:
     content: "{{ lookup('file', 'settings1c.reg') }}"
   register: merge12_result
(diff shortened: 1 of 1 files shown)


plugins/modules/win_regmerge.ps1
```diff
@@ -28,15 +28,33 @@ $result = @{
...
+
+if (  $path -and $content ) {
+    Fail-Json -obj $result -message "Extra arguments: path or content. Only one must be provided."
+}
+
+if ( $content ) {
+    $guid = [guid]::NewGuid()
+    $path = $env:TEMP + "\" + $guid.ToString() + 'ansible_win_regmerge_content.reg'
+
+    $content | Out-File -FilePath $path
```
jborean93, line 47: Looks like `reg.exe IMPORT` doesn't care about the extension so we can simply do

```suggestion
    $path = [System.IO.Path]::GetTempFileName()
    Set-Content -LiteralPath $path -Value $content
```

plugins/modules/win_regmerge.py
```diff
@@ -21,8 +21,11 @@
     description:
       - The full path including file name to the registry file on the remote machine to be merged
     type: path
-    required: yes
-  compare_key:
+  content:
+    description:
+      - When used instead of O(path), merges the value specified into the Windows registry
+    type: str
```
jborean93, line 27: We need to document when the new parameter was added.

```suggestion
    type: str
    version_added: 2.2.0
```

jborean93 (changes requested):
We would need to a changelog fragment for this change https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment. This would be documented under the `minor_changes` key.

It would also be great to add an integration test for this scenario under https://github.com/ansible-collections/community.windows/blob/main/tests/integration/targets/win_regmerge/tasks/main.yml. We could simply change the `win_regmerge` entries under https://github.com/ansible-collections/community.windows/blob/bd57fc1457b3183eb299effaeac7576e6e04f8e4/tests/integration/targets/win_regmerge/tasks/main.yml#L65-L85 to use `content: '{{ lookup("file", "settings2.reg") }}'` instead of `path` there.

jborean93 (approved):

Issue #410, opened 2022-06-20
win_iis_webapppool cant set attributes for pool with "value" in the name


timothyanderson82:
##### SUMMARY

win_iis_webapppool module fails with the below error when the word "value" in the app pool name and you try and set attributes:

```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: at <ScriptBlock>, <No file>: line 225
fatal: [hob.hey.net]: FAILED! => {"changed": false, "msg": "Unhandled exception while executing module: You cannot call a method on a null-valued expression."}
```

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

community.windows.win_iis_webapppool

##### ANSIBLE VERSION

```paste below
[DEPRECATION WARNING]: Ansible will require Python 3.8 or newer on the controller starting with Ansible 2.12. Current version: 3.6.8 (default, Nov 16 2020, 16:55:22) [GCC 4.8.5 20150623 (Red Hat 
4.8.5-44)]. This feature will be removed from ansible-core in version 2.12. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
/usr/local/lib/python3.6/site-packages/ansible/parsing/vault/__init__.py:44: CryptographyDeprecationWarning: Python 3.6 is no longer supported by the Python core team. Therefore, support for it is deprecated in cryptography and will be removed in a future release.
  from cryptography.exceptions import InvalidSignature
ansible [core 2.11.12] 
  config file = None
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.6/site-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/bin/ansible
  python version = 3.6.8 (default, Nov 16 2020, 16:55:22) [GCC 4.8.5 20150623 (Red Hat 4.8.5-44)]
  jinja version = 3.0.3
  libyaml = True
```

##### COLLECTION VERSION

```paste below
[DEPRECATION WARNING]: Ansible will require Python 3.8 or newer on the controller starting with Ansible 2.12. Current version: 3.6.8 (default, Nov 16 2020, 16:55:22) [GCC 4.8.5 20150623 (Red Hat 
4.8.5-44)]. This feature will be removed from ansible-core in version 2.12. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
[...]

PR #587, opened 2024-10-30
fixing portion where building app pools with the word value fails.

AnsibleJosh:
##### SUMMARY

For the last two years we've had an issue with this module regard app pool creations failing that specifically contain the word "value" in it. We weren't able to create sites that contained this word. 

**https://github.com/ansible-collections/community.windows/issues/410**

Root cause is that the string "Value" is matched against collection of property objects of the $attribute_value but from the context it can be deduced that "Value" should only be matched against collection of names of those properties. Match should not be partial either, hence the -eq operator.

##### ISSUE TYPE

- Bugfix Pull Request
-
##### COMPONENT NAME

community.win_iis_webapppool

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/588-win_iis_webapppool.yaml
+++ b/changelogs/fragments/588-win_iis_webapppool.yaml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+  - win_iis_webapppool  - this pull request fixes the portion where building an app pool with the word "value" in it fails unexpectedly.
+    https://github.com/ansible-collections/community.windows/issues/410.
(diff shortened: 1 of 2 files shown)


jborean93 (commented):
Thanks for the bugfix, is it possible to add a changelog fragment to document this bugfix https://docs.ansible.com/ansible/latest/community/development_process.html#creating-changelog-fragments

jborean93:
Basically create a file in the [changelogs/fragments](https://github.com/ansible-collections/community.windows/tree/main/changelogs/fragments) directory with something like

```yaml
bugfixes:
  - >-
    win_iis_webapppool - Brief description of what is fixed
    - https://github.com/ansible-collections/community.windows/issues/410
```

jborean93 (approved):

jborean93:
Thanks for the bugfix!

Issue #147, opened 2020-10-06
win_psmodule fails due to PackageManagement requiring NuGet


mjiderhamn:
##### SUMMARY

`win_psmodule` fails due to `PackageManagement` requiring NuGet.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

`win_psmodule`

##### ANSIBLE VERSION

```paste below
ansible 2.10.2
  config file = /cygdrive/l/sandbox/ip-infrastructure/ansible/ansible.cfg
  configured module search path = ['/home/redacted/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.6/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.6.10 (default, Jun  8 2020, 19:37:31) [GCC 9.3.0]
```

##### CONFIGURATION

```paste below
ANSIBLE_SSH_ARGS(/cygdrive/redacted/ansible/ansible.cfg) = -o ForwardAgent=yes -o ControlMaster=no -o PasswordAuthentication=no
```

##### OS / ENVIRONMENT

Windows Server 2012 R2 Datacenter
PowerShell 5.1.14409.1018

##### STEPS TO REPRODUCE

Apply playbook

```yaml
- name: Install PSCX
  win_psmodule:
    name: Pscx
    allow_clobber: yes
```

##### EXPECTED RESULTS

Module is installed

##### ACTUAL RESULTS

`win_psmodule.ps1` has an `Install-NugetProvider` function to make sure NuGet provider is available, however it is called only **after** `Import-Module -Name PackageManagement, PowerShellGet` at [line 427](https://github.com/ansible-collections/community.windows/blob/main/plugins/modules/win_psmodule.ps1#L427).

```paste below
fatal: [my-host]: FAILED! => {
    "changed": false,
[...]

PR #588, opened 2024-11-04
Issue #147

mfortin:
##### SUMMARY
Attempting to fix Issue #147 as a confirmation is required when it should not.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
`win_psmodule`

##### ADDITIONAL INFORMATION

Before:
```
The full traceback is:
Exception calling "ShouldContinue" with "2" argument(s): "A command that prompts the user failed because the host program or the command type does not support user interaction. The host was attempting to request confirmation with the following message: PowerShellGet requires NuGet provider version '2.8.5.201' or newer to interact with NuGet-based repositories. The NuGet provider must be available in 'C:\Program Files\PackageManagement\ProviderAssemblies' or
'C:\Users\user\AppData\Local\PackageManagement\ProviderAssemblies'. You can also install the NuGet provider by running 'Install-PackageProvider -Name NuGet -MinimumVersion 2.8.5.201 -Force'. Do you want PowerShellGet to install and import the NuGet provider now?"
At line:131 char:13
+             $null = $job | Receive-Job -AutoRemoveJob -Wait -ErrorAct ...
+             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Install-NuGetClientBinaries], MethodInvocationException
    + FullyQualifiedErrorId : HostException,Install-NuGetClientBinaries

at Install-NuGetClientBinaries, C:\Program Files\WindowsPowerShell\Modules\PowerShellGet\1.0.0.1\PSModule.psm1: line 7392
at Install-Module, C:\Program Files\WindowsPowerShell\Modules\PowerShellGet\1.0.0.1\PSModule.psm1: line 1725
at , : line 26
fatal: [10.0.0.11]: FAILED! => {
    "changed": true,
    "msg": "Problems adding a prerequisite module PackageManagement or PowerShellGet: Exception calling \"ShouldContinue\" with \"2\" argument(s): \"A command that prompts the user failed because the host program or the command type does not support user interaction. The host was attempting to request confirmation with the following message: PowerShellGet requires NuGet provider version '2.8.5.201' or newer to interact with NuGet-based repositories. The NuGet provider must be available in 'C:\\Program Files\\PackageManagement\\ProviderAssemblies' or 'C:\\Users\\user\\AppData\\Local\\PackageManagement\\ProviderAssemblies'. You can also install the NuGet provider by running 'Install-PackageProvider -Name NuGet -MinimumVersion 2.8.5.201 -Force'. Do you want PowerShellGet to install and import the NuGet provider now?\"",
    "nuget_changed": false,
    "output": "",
    "repository_changed": false
}
```
After
```
changed: [10.0.0.11] => {
    "changed": true,
    "nuget_changed": false,
    "output": "Module Microsoft.PowerShell.PSResourceGet installed",
    "repository_changed": false
}
```

### DIFF
--- a/changelogs/fragments/588-nuget-requirement-patch.yaml
+++ b/changelogs/fragments/588-nuget-requirement-patch.yaml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - This issue fixes installation of requirements as it requires a confirmation when installed as a depedency to PowershellGet. Installing it by itself prevents this confirmation dialog and allows required components to be installed (https://github.com/ansible-collections/community.windows/issues/147).
--- a/changelogs/fragments/588-nuget-requirement-patch.yaml
+++ b/changelogs/fragments/588-nuget-requirement-patch.yaml
@@ -1,3 +1,5 @@
 ---
 minor_changes:
-  - This issue fixes installation of requirements as it requires a confirmation when installed as a depedency to PowershellGet. Installing it by itself prevents this confirmation dialog and allows required components to be installed (https://github.com/ansible-collections/community.windows/issues/147).
+  - This issue fixes installation of requirements as it requires a confirmation when installed as a depedency to PowershellGet.
+    Installing it by itself prevents this confirmation dialog and allows required components to be installed
+    (https://github.com/ansible-collections/community.windows/issues/147).
--- a/plugins/modules/win_psmodule.ps1
+++ b/plugins/modules/win_psmodule.ps1
@@ -101,7 +101,7 @@ Function Install-PrereqModule {
         $job = Start-Job -ScriptBlock {
             $ErrorActionPreference = 'Stop'
 
-            if (-not (Get-PackageProvider -ListAvailable -ErrorAction SilentlyContinue | Where-Object { ($_.name -eq 'Nuget') -and ($_.version -ge "2.8.5.201") }) ) {
+            if (-not (Get-PackageProvider -ListAvailable | Where-Object { ($_.name -eq 'Nuget') -and ($_.version -ge "2.8.5.201") }) ) {
                 Install-PackageProvider -Name NuGet -MinimumVersion 2.8.5.201 -Force | Out-Null
             }
 
--- a/plugins/modules/win_psmodule.ps1
+++ b/plugins/modules/win_psmodule.ps1
@@ -70,6 +70,9 @@ Function Install-PrereqModule {
         [string]$Repository
     )
 
+    # Install NuGet provider if needed.
+    Install-NugetProvider -CheckMode $CheckMode
+
     # Those are minimum required versions of modules.
     $PrereqModules = @{
         PackageManagement = '1.1.7'
@@ -108,7 +111,6 @@ Function Install-PrereqModule {
                     Name = $info.Name
                     MinimumVersion = $info.Version
                     Force = $true
-                    Confirm = $false
                 }
                 $installCmd = Get-Command -Name Install-Module
                 if ($installCmd.Parameters.ContainsKey('SkipPublisherCheck')) {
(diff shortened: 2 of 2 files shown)


jborean93:
Thanks for the PR, are you able to create a changelog fragment for this bugfix Thanks for the bugfix, is it possible to add a changelog fragment to document this bugfix https://docs.ansible.com/ansible/latest/community/development_process.html#creating-changelog-fragments.

Just to confirm this fixes a problem when running in check mode and the nuget provider needed to be installed?

mfortin:
@jborean93 I added the fragment.

The problem also occurs when not running in check mode.
`Install-PrereqModule` function checks and installs `PackageManagement` and `PowerShellGet`. However, it doesn't check if Nuget package provider is installed prior to this installation. If it isn't installed, there is a confirmation dialog that waits for user input to confirm installation of this other dependency. By running `Install-NugetProvider` as part of the `Install-PrereqModule`, it no longer is a dependency of another package and can be automatically accepted and installation of the module succeeds.

jborean93:
Sorry I misread the change and thought it was just adding `-CheckMode`. We'll have to move the nuget provider pre-req check and install into the `Start-Job` below https://github.com/ansible-collections/community.windows/blob/main/plugins/modules/win_psmodule.ps1#L104. Unfortunately we cannot call any `PackageManagement` or `PowerShellGet` functions until the actual pre-reqs have been satisfied. If we were to call any `PackageManagement` function before this we could be inadvertently loading an older version of either package causing problems later on when trying to use the newer one.

Keep in mind the code inside the `Start-Job` is standalone and won't have access to any of the functions defined in the module.

mfortin:
I tested it and it works out just fine.

```Install-PackageProvider -Name NuGet -MinimumVersion 2.8.5.201 -Force -WhatIf:$CheckMode | Out-Null```
Needs to be done before the Start-Job, or at least before ```Install-Module PackageManagement```.

**Comment by jborean93** [comment]
> I tested it and it works out just fine.

The issue that goes into more detail for this is https://github.com/ansible-collections/community.windows/issues/487 but I'll try to explain it again.

It's a complicated scenario that doesn't happen for all scenarios but for some we still need to handle. The problem is using any of the cmdlets from `PackageManagement` before we've checked to ensure we have our minimum requirement will cause PowerShell to load that older version in the process and we cannot unload or load the newer one. This is ok for some situations but others where `Install-Module` might pass through extra arguments to `PackageManagement` during an install and if `PackageManagement` isn't new enough to understand these parameters it fails. IIRC this was seen with the `-AcceptLicense` parameter. To compat this we put all the pre-requisite work into a sub-process that we are spawning through `Start-Job`. Through this we can ensure that `PackageManagement` and `PowerShellGet` are installed to the minimums we require when we load it into the main module process.

In the case you are trying to fix, calling `Install-PackageProvider` before we have ensured `PackageManagement` is up to date we will load that older version which doesn't support things like `-AcceptLicense`. We need to do the checks for `Install-NugetProvider` into the job as well. The simplest solution is to add the following to the top of the job's actions

```powershell
$nugetProvider = Get-PackageProvider -ListAvailable | Where-Object { ($_.name -eq 'Nuget') -and ($_.version -ge "2.8.5.201") }
if (-not($nugetProvider)) {
    Install-PackageProvider -Name NuGet -MinimumVersion 2.8.5.201 -Force | Out-Null
}
```

We don't need to worry about check mode here because the job won't be run in check mode.

jborean93 (approved):
Thanks for working on this issue, appreciate the patience while I try to explain the current way the module works.

PR #595, opened 2024-12-04
New Module: win_net_tcp, win_net_sock

Yannnyan:
###### Rational
Sometimes network administrators if I can consider myself as one, would like to test their firewall products with ansible using connections on specific ports.
###### Current state
ATM running powershell is the best option.
###### Features
These two modules provides easy way to create listening socket and connect to it.</br>Providing enough details to debug whether the packets are being dropped, or the socket is not listening, or the connection has succeeded.</br> Nevertheless you can stop processes running on specific ports with win_net_sock.</br>
###### Two modules
I chose to push two modules at once, because one module helps debug the other.
###### Design choices
Using WmiObject for creating a detached process is the best way i found on windows to provide a running socket that does not close when the session closes nor when the parent process exits.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

win_net_tcp
win_net_sock
##### ADDITIONAL INFORMATION

Before using this module you would have to run in powershell or win_shell these commands:
- Create script to create listener and start listening
- Create detached process to have to execute the script

- Create a connection through a socket on another host
- Interpret the error messages of powershell via result

```paste below

```


jborean93:
Hi, thank you for your contribution but at this point in time we are not accepting new modules for this collection as per our contribution section in the README https://github.com/ansible-collections/community.windows/tree/main?tab=readme-ov-file#contributing-to-this-collection.

I know you've put in some effort to create these modules and open this PR it's just at this time we have limited resources to maintain this collection and adding new plugins increases that burden even more.

PR #599, opened 2025-01-10
Add force option to win psrepository copy

tgoetheyn:
##### SUMMARY
Adds a `force` option to the **win_psrepository_copy** module.
The current module only had the option to append. This addition also lets you 'set' the repository list identically tot the source

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
win_psrepository_copy

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/win_psrepository_copy.py
+++ b/plugins/modules/win_psrepository_copy.py
@@ -55,6 +55,11 @@ options:
       - systemprofile
       - LocalService
       - NetworkService
+  force:
+    description:
+      - If C(True), this deletes repositories that are not present in the source.
+    type: bool
+    default: False
 notes:
   - Does not require the C(PowerShellGet) module or any other external dependencies.
   - User profiles are loaded from the registry. If a given path does not exist (like if the profile directory was deleted), it is silently skipped.
--- a/plugins/modules/win_psrepository_copy.ps1
+++ b/plugins/modules/win_psrepository_copy.ps1
@@ -39,6 +39,10 @@ $spec = @{
                 'NetworkService'
             )
         }
+        force = @{
+            type = 'bool'
+            default = $false
+        }
     }
 }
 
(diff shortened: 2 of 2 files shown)


plugins/modules/win_psrepository_copy.py
```diff
@@ -55,6 +55,11 @@
       - systemprofile
       - LocalService
       - NetworkService
+  force:
+    description:
+      - If C(True), this deletes repositories that are not present in the source.
+    type: bool
+    default: False
```
jborean93, line 62: We need to include the version when this new option was first present in which will be 2.4.0.

```suggestion
    default: False
    version_added: 2.4.0
```

jborean93 (changes requested):
Thanks for the PR, there are two more things we would need to see;

+ A changelog fragment documenting this new feature - https://docs.ansible.com/ansible/latest/community/development_process.html#creating-changelog-fragments
+ An integration test added to make sure this new feature is idempotent and does what it should be doing

The current integration tests are located https://github.com/ansible-collections/community.windows/tree/main/tests/integration/targets/win_psrepository_copy/tasks. I think adding a new `test_force_profiles.yml` and importing it at https://github.com/ansible-collections/community.windows/blob/237af1a0a5b961c335b5d194bdcea2be3d6ffb8d/tests/integration/targets/win_psrepository_copy/tasks/main.yml#L77 would be ideal. The `test_force_profiles.yml` file would be like [test_include_repo.yml](https://github.com/ansible-collections/community.windows/blob/main/tests/integration/targets/win_psrepository_copy/tasks/test_include_repo.yml) but with the `force` option and updated expectation at the bottom.

tgoetheyn:
Hi @jborean93 
I added the changelog and wrote an integration test.

Everything looks fine, but the pipeline keeps failing on "Ansible devel Lint" and I can't figure out why?

jborean93:
Looks like `ansible-lint` in CI is ignoring our ignore rules file, I'll need to look into that separately so we can skip that check for this PR. Appreciate you working on this.

jborean93 (approved):

PR #600, opened 2025-01-16
Fix up ansible-lint version

jborean93:
##### SUMMARY
Bumps `ansible-lint` to the version used by galaxy importer in production. Also pins `ansible-compat` due to an incompatibility with the latest release. This will be removed in the future once that bug is fixed.

##### ISSUE TYPE
- Bugfix Pull Request


jborean93:
CI error was not a failure in the tests but rather the publishing of the test results which is unrelated to the change here.

PR #606, opened 2025-02-05
Fix mismatch of return structure in win_file_version module documentation

easymoney322:
Wrapped return values in a complex dictionary so format in docs matches the real result. Changed the order to match the order in output.
Added a sample for one of the parameters

##### SUMMARY

Fixes #605 

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

win_file_version
##### ADDITIONAL INFORMATION

The only thing that was changed is the documentation file to match the style of [win_stat module](https://docs.ansible.com/ansible/latest/collections/ansible/windows/win_stat_module.html).

PR #607, opened 2025-02-10
Add file_version_raw result value to win_file_version

easymoney322:
##### SUMMARY

I suggest adding return value `FileVersionRaw`, as it behaves differently from the `FileVersion`.
`FileVersionRaw` output is guaranteed to be a non-empty string that has strict format.
The `FileVersionRaw` might match the version, be older, newer or be set to `0.0.0.0`.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

win_file_version
##### ADDITIONAL INFORMATION

There are various posts that demonstrate difference in these version values.
Some are on [StackOverflow](https://stackoverflow.com/a/66005), some are on [GitHub](https://github.com/notepad-plus-plus/notepad-plus-plus/issues/13842). The thing is, these versions may match or differ.

In my case, the `FileVersion` was missing but the `FileVersionRaw` wasn't and the [windows explorer was showing the raw version](https://imgur.com/4HkR8Jf). In some cases it's better to use the raw version, in some cases it's the opposite.

Thus, both `file_version` and `file_version_raw` are needed. 
Below are two examples that demonstrate it:

The task
```
- name: Get acm instance version
  win_file_version:
    path: 'C:\\Program Files//windows_exporter/windows_exporter.exe'
  register: exe_file_version

- debug:
    msg: '{{ exe_file_version }}'
```

Before:
```paste below
ok: [desktop-rpseca1] => {
    "msg": {
        "changed": false,
        "failed": false,
        "win_file_version": {
            "file_build_part": "1",
            "file_major_part": "0",
            "file_minor_part": "25",
            "file_private_part": "0",
            "file_version": "",
            "path": "C:\\\\Program Files//windows_exporter/windows_exporter.exe",
            "product_version": ""
        }
    }
}
```

After:
```paste below
TASK [debug] ****************************************************************************************************************************
task path: /etc/ansible/playbooks/version.yml:6
ok: [desktop-rpseca1] => {
    "msg": {
        "changed": false,
        "failed": false,
        "win_file_version": {
            "file_build_part": "1",
            "file_major_part": "0",
            "file_minor_part": "25",
            "file_private_part": "0",
            "file_version": "",
            "file_version_raw": "0.25.1.0",
            "path": "C:\\\\Program Files//windows_exporter/windows_exporter.exe",
            "product_version": ""
        }
    }
}
```
Notice that `file_version` is still empty.

This may also help for string comparison when the developers change `file_version` so it starts to return a value with 'v' prefix:
```
ok: [desktop-rpseca1] => {
    "msg": {
        "changed": false,
        "failed": false,
        "win_file_version": {
            "file_build_part": "2",
            "file_major_part": "0",
            "file_minor_part": "30",
            "file_private_part": "0",
            "file_version": "v0.30.2",
            "file_version_raw": "0.30.2.0",
            "path": "C:\\\\Program Files//windows_exporter/windows_exporter.exe",
            "product_version": "v0.30.2"
        }
    }
}
```

Since docs had to be updated, I've also removed `changed` return value as it's both [wasn't necessary](https://github.com/ansible-collections/community.windows/pull/606#issuecomment-2637700614) and [hardcoded to return `false` anyway](https://github.com/ansible-collections/community.windows/blob/bc8260613daa71873f5101147eefafc300e72e21/plugins/modules/win_file_version.ps1#L12).

### DIFF
--- a/plugins/modules/win_file_version.ps1
+++ b/plugins/modules/win_file_version.ps1
@@ -48,7 +48,7 @@ Try {
     If ($null -eq $file_private_part) {
         $file_private_part = ''
     }
-	$file_version_raw = $_version_fields.FileVersionRaw
+    $file_version_raw = $_version_fields.FileVersionRaw
     If ($null -eq $file_version_raw) {
         $file_version_raw = ''
     }
--- a/plugins/modules/win_file_version.ps1
+++ b/plugins/modules/win_file_version.ps1
@@ -19,7 +19,7 @@ If (-Not (Test-Path -LiteralPath $path -PathType Leaf)) {
 }
 $ext = [System.IO.Path]::GetExtension($path)
 If ( $ext -notin '.exe', '.dll') {
-    Fail-Json $result "Specified path $path is not a valid file type; Must be DLL or EXE."
+    Fail-Json $result "Specified path $path is not a valid file type; must be DLL or EXE."
 }
 
 Try {
--- a/plugins/modules/win_file_version.ps1
+++ b/plugins/modules/win_file_version.ps1
@@ -19,7 +19,7 @@ If (-Not (Test-Path -LiteralPath $path -PathType Leaf)) {
 }
 $ext = [System.IO.Path]::GetExtension($path)
 If ( $ext -notin '.exe', '.dll') {
-    Fail-Json $result "Specified path $path is not a valid file type; must be DLL or EXE."
+    Fail-Json $result "Specified path $path is not a valid file type; Must be DLL or EXE."
 }
 
 Try {
@@ -48,6 +48,10 @@ Try {
     If ($null -eq $file_private_part) {
         $file_private_part = ''
     }
+	$file_version_raw = $_version_fields.FileVersionRaw
+    If ($null -eq $file_version_raw) {
+        $file_version_raw = ''
+    }
 }
 Catch {
     Fail-Json $result "Error: $_.Exception.Message"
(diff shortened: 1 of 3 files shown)


plugins/modules/win_file_version.py
```diff
@@ -32,15 +33,10 @@
   register: exe_file_version
 
 - debug:
-    msg: '{{ exe_file_version }}'
+    msg: '{{ exe_file_version }}.win_file_version'
```
jborean93, line 36: The sub key needs to be in the template.

```suggestion
    msg: '{{ exe_file_version.win_file_version }}'
```

plugins/modules/win_file_version.ps1
```diff
@@ -19,7 +19,7 @@ If (-Not (Test-Path -LiteralPath $path -PathType Leaf)) {
 }
 $ext = [System.IO.Path]::GetExtension($path)
 If ( $ext -notin '.exe', '.dll') {
-    Fail-Json $result "Specified path $path is not a valid file type; must be DLL or EXE."
+    Fail-Json $result "Specified path $path is not a valid file type; Must be DLL or EXE."
```
jborean93, line 22: While I doubt someone is checking this string it is best to keep it as before. This is so someone doing the below will continue to work

```yaml
- community.windows.win_file_version:
    path: ...
  register: file_info
  failed_when:
  - file_info is failed
  - not file_info.msg is search('.* must be DLL or EXE.*')
```

```suggestion
    Fail-Json $result "Specified path $path is not a valid file type; must be DLL or EXE."
```

jborean93 (changes requested):

easymoney322:
Hello,
I'm sorry I don't know how could I possibly make a mistake this big with the template. Thanks for noticing!

I too think compatibility should be prioritized over grammar. I've reverted that change as you suggested.

jborean93:
That's all good, I make mistakes all the time. Thanks for submitting the PR and changelog fragment.

PR #615, opened 2025-03-14
fix logic in check loop

Rogacz:
##### SUMMARY
Fix logic in loop to check options for potential failed state in case of force=false parameter.

Currently, checks are not performed and **$pristine** is left unset, causing failure.

```
The full traceback is:
The variable '$pristine' cannot be retrieved because it has not been set.
At line:213 char:9
+     if ($pristine) {
+         ~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (pristine:String) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : VariableIsUndefined

ScriptStackTrace:
at <ScriptBlock>, <No file>: line 213
fatal: [hostname]: FAILED! => {
    "changed": false,
    "msg": "Unhandled exception while executing module: The variable '$pristine' cannot be retrieved because it has not been set."
}


```

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
win_format

##### ADDITIONAL INFORMATION
As I understand, the loop in lines 171-194 is supposed to run checks for a given partition when the access path is matching. 
Currently, it checks for not matching path and for example in case of newly created single partition it never runs content of if block in line 172 and never sets **$pristine** flag causing fail.

The bug never happens when force flag is set to true.


Rogacz:
Well now it fails test cause $Path is empty in your test cases as you use drive_letter
I was using $path in my tests and didn't catch it.

`    if ($access_path -eq $Path || $access_path[0].toUpper() -eq $drive_letter.toUpper()) {`
This should work for line 172 for this case but will fail on running module with Label 

I will think about something unless anyone has suggestion :)

Rogacz:
This fix is completed and can be merged. Do I need to add or change anything?

jborean93:
Sorry for missing this PR, would you be able to create a changelog fragment for this `bugfix` please. Docs on how to do this can be found under https://docs.ansible.com/ansible/latest/community/development_process.html#creating-changelog-fragments.

jborean93:
Thanks for the bugfix and once again sorry for the delay in reviewing this.

Issue #631, opened 2025-05-07
win_xml type: attribute with state: absent not functioning as intended


owadestig:
##### SUMMARY
win_xml remove attribute broken. Intended behavior:

`- name: Remove attribute
  community.windows.win_xml:
    path: some_path
    xpath: /some/xpath/some_element
    state: absent
    type: attribute
    attribute: some_attribute`

Should remove attribute `some_attribute` if the element `some_element` has it. Will reach else block in `win_xml.ps1` starting on line 266. Here it fails the if statement `if ($node.Name -eq $attribute ) `. If statement should be `if ($node.HasAttribute($attribute))` instead.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
win_xml.ps1 

##### ANSIBLE VERSION
```
ansible [core 2.15.9]
```

##### COLLECTION VERSION
```
community.windows 2.3.0
```

##### CONFIGURATION

```paste below

```

##### OS / ENVIRONMENT

windows virtual machine

##### STEPS TO REPRODUCE

See bug description.

```yaml

```

##### EXPECTED RESULTS
attribute removed

##### ACTUAL RESULTS
attribute not removed. No error logged. No description of what happened, code doesn't cover this specific case.

Akasurde:
@owadestig Are you still interested in fixing this issue? Thanks.

PR #637, opened 2025-06-30
win_xml: Handle attribute removal correctly

Akasurde:
##### SUMMARY

Fixes: #631

Signed-off-by: Abhijeet Kasurde <Akasurde@redhat.com>

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
changelogs/fragments/win_xml_attribute.yml
plugins/modules/win_xml.ps1
tests/integration/targets/win_xml/tasks/main.yml

### DIFF
--- a/tests/integration/targets/win_xml/tasks/main.yml
+++ b/tests/integration/targets/win_xml/tasks/main.yml
@@ -1,20 +1,7 @@
 # test code for the Windows xml module
-# (c) 2017, Richard Levenberg <richard.levenberg@cosocloud.com>
-
-# This file is part of Ansible
-#
-# Ansible is free software: you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published by
-# the Free Software Foundation, either version 3 of the License, or
-# (at your option) any later version.
-#
-# Ansible is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


tests/integration/targets/win_xml/tasks/main.yml
```diff
@@ -100,6 +87,60 @@
...
+  ansible.windows.win_stat:
+    path: "{{ remote_tmp_dir }}\\config.xml"
+    get_checksum: yes
+  register: config_xml_checksum_after
+
+- name: check attribute is removed
+  assert:
+    that:
+    - attribute_remove_result is changed
+    - config_xml_checksum_before.stat.checksum != config_xml_checksum_after.stat.checksum
```
jborean93, line 115: This just verifies the checksum has changed, we should be checking that the relevant attribute was removed otherwise anything could have changed.

jborean93: I would either just return the content through `slurp` or a `win_powershell` calling `Get-Content ... -Raw`. Or have a `win_powershell` inline task that validates the expected structure of the XML content, e.g. the attribute is not present.

PR #665, opened 2025-12-20
win_pssession_configuration: Fix types for parameters passed in New-PSSessionConfigurationFile

tortenairbag:
##### SUMMARY

Since Ansible 12, the module `win_pssession_configuration` fails if these parameters are used: `ModulesToImport`, `VisibleCmdlets`, `VisibleFunctions`
The error is always:
> The member 'ModulesToImport' must be an array consisting of either string or hashtable elements.

This error message from the PS CmdLet is also misleading; invalid values for the parameters `VisibleCmdlets` and `VisibleFunctions` can cause this too, even though `ModulesToImport` is explicitly blamed as cause.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

win_pssession_configuration

##### ADDITIONAL INFORMATION

This bug is caused by changes made in the ansible-core; the same module script works fine with Ansible 11 and below.
After some debugging in the ansible-core repo, I identified that ansible/ansible#84868 caused this issue.
```
git bisect good
7d0886457ac7d84219583115f131f67d95eb9987 is the first bad commit
commit 7d0886457ac7d84219583115f131f67d95eb9987
Author: Jordan Borean <jborean93@gmail.com>
Date:   Tue Apr 1 11:24:35 2025 +1000

    Windows Exec Runner Update (#84868)
    
    Updates the Windows exec runner...
```

This change introduced some major changes on how the module parameters are parsed and passed to the module.
A valid approach would be to fix this compatibility bug in the ansible-core repo instead in this module only, but I decided against it because:
- I assume this issue arises only if the parameter spec uses `type = 'raw'`, that is commonly used and seems to cause no issues in the other plugins. Excerpt about assumed usages, in scope are only the `ansible.windows` and `community.windows` collections:
  ```
  $ egrep -Hirn "[\"']raw['\"]"
  ansible/windows/plugins/modules/setup.ps1:34:                           $spec.options.$prop = @{ type = 'raw' }
  ansible/windows/plugins/modules/setup.ps1:45:                           $spec.options."$($param.Key)" = @{ type = 'raw' }
  ansible/windows/plugins/modules/win_dsc.ps1:36:                         Unknown { "raw" }
  ansible/windows/plugins/modules/win_dsc.ps1:37:                         default { "raw" }
  ansible/windows/plugins/modules/win_http_proxy.ps1:12:                  proxy = @{ type = "raw" }
  ansible/windows/plugins/modules/win_inet_proxy.ps1:13:                  proxy = @{ type = "raw" }
  ansible/windows/plugins/modules/win_package.ps1:1281:                   arguments = @{ type = "raw" }
  ansible/windows/plugins/modules/win_regedit.ps1:16:                     data = @{ type = 'raw' }
  ansible/windows/plugins/modules/win_uri.ps1:17:                         body = @{ type = "raw" }
  community/windows/plugins/modules/win_http_proxy.ps1:12:                proxy = @{ type = "raw" }
  community/windows/plugins/modules/win_inet_proxy.ps1:13:                proxy = @{ type = "raw" }
  community/windows/plugins/modules/win_pssession_configuration.ps1:44:   modules_to_import = @{ type = 'list' ; elements = 'raw' }
  community/windows/plugins/modules/win_pssession_configuration.ps1:46:   visible_cmdlets = @{ type = 'list' ; elements = 'raw' }
  community/windows/plugins/modules/win_pssession_configuration.ps1:47:   visible_functions = @{ type = 'list' ; elements = 'raw' }
  ```
- I assume in most modules the ansible parameters are not simply passed to a PS CmdLets as-is but always transformed with some custom logic first.
  - `win_http_proxy`/`win_inet_proxy`: Type checks for `$proxy` for proper validation before serializing it as string
  - `win_package`: Transformed with `ConvertTo-EscapedArgument` and always serialized to string
  - `win_regedit`: Complex type checks and proper payload casting before passing it to 
  - `win_uri`: Always serialized as string and passed as buffer with `[System.Text.Encoding]::UTF8.GetBytes($body_string)`
- Especially challenging in this scenario is that complex and nested structures can be passed for the `VisibleCmdlets`/`VisibleFunctions` parameters, as shown in my example below. A simple serialization-based normalization like `$payload | ConvertTo-Json -Depth 99 -Compress | ConvertFrom-Json` with PS-Native CmdLets did not properly transform the objects in compatible types.
- Last but not least, whoever wants to inspect this issue in future: Adding some debug messages to report the object types like "Type VisibleCmdlets: $($ParameterSet.VisibleCmdlets.GetType().FullName)" is misleading. For both Ansible 11 and Ansible 12 the reported types were identical - for all nested members too - but somehow the objects passed by Ansible 12 were rejected by `New-PSSessionConfigurationFile` (Debugging this issue was real fun, I can tell y'all ^^).

Playbook:
```
  - name: Create a complex configuration
    community.windows.win_pssession_configuration:
      name: "{{ config_name }}"
      description: "{{ config_description }}"
      #modules_to_import: Microsoft.PowerShell.Utility  # <= No value passed, even though error message blames it
      access_mode: remote
      execution_policy: restricted
      language_mode: no_language
      visible_external_commands: []
      visible_functions:
        - tabexpansion2
        - help
      visible_cmdlets:
        - Get-Help
        - Name: Get-Service
          Parameters:
            - Name: DependentServices
            - Name: RequiredServices
            - Name: Name
              ValidateSet:
                - WinRM
                - W3SVC
                - WAS
      visible_aliases:
        - gsv
      security_descriptor_sddl: "O:NSG:BAD:P(A;;GA;;;IU)(A;;GA;;;BA)(A;;GA;;;RM)S:P(AU;FA;GA;;;WD)(AU;SA;GXGW;;;WD)"
    register: status
```
Output:
```
[ERROR]: Task failed: Unhandled exception while executing module: The member 'ModulesToImport' must be an array consisting of either string or hashtable elements.
Origin: /xxx/play.yml:13:5

11       state: absent
12
13   - name: Create a complex configuration
       ^ column 5

fatal: [LenovoWin]: FAILED! => {"changed": false, "msg": "Task failed: Unhandled exception while executing module: The member 'ModulesToImport' must be an array consisting of either string or hashtable elements."}
```

### DIFF
--- a/plugins/modules/win_pssession_configuration.ps1
+++ b/plugins/modules/win_pssession_configuration.ps1
@@ -223,7 +223,7 @@ function ConvertTo-PsNative {
     #>
     [CmdletBinding()]
     param(
-        [Parameter(Mandatory=$true, ValueFromPipeline=$true)]
+        [Parameter(Mandatory = $true)]
         $Value
     )
 
@@ -231,7 +231,7 @@ function ConvertTo-PsNative {
         if ($null -eq $Value) { return $null }
 
         # Unwrap PSObject wrapper if present
-        if ($Value -is [System.Management.Automation.PSObject] -and $Value.BaseObject -ne $null) {
+        if ($Value -is [System.Management.Automation.PSObject] -and $null -ne $Value.BaseObject) {
             $Value = $Value.BaseObject
         }
 
@@ -301,7 +301,8 @@ function Write-GeneratedSessionConfiguration {
         foreach ($k in $ParameterSet.Keys) {
             if ($k -in ('ModulesToImport', 'VisibleCmdlets', 'VisibleFunctions')) {
                 $p[$k] = ConvertTo-PsNative $ParameterSet[$k]
-            } else {
+            }
+            else {
                 $p[$k] = $ParameterSet[$k]
             }
         }
--- a/changelogs/fragments/665-win_pssession_configuration_types.yml
+++ b/changelogs/fragments/665-win_pssession_configuration_types.yml
@@ -1,3 +1,2 @@
 bugfixes:
 - 'win_pssession_configuration - Fix type errors for parameters passed in New-PSSessionConfigurationFile'
-


plugins/modules/win_pssession_configuration.ps1
```diff
@@ -203,6 +203,155 @@ function ConvertFrom-AnsibleOption {
...
+
+        if ($valueType.IsGenericType -and $valueType.GetGenericTypeDefinition().FullName -eq 'System.Collections.Generic.List`1') {
+            foreach ($e in $Value) { $out += (ConvertTo-PsNative -Value $e) }
+        }
+        elseif ($valueType.FullName -eq 'System.Collections.ArrayList') {
+            foreach ($e in $Value) { $out += (ConvertTo-PsNative -Value $e) }
+        }
+        elseif ($valueType.IsArray) {
+            foreach ($e in $Value) { $out += (ConvertTo-PsNative -Value $e) }
+        }
```
jborean93: All these types above this inherit `IList` so it should be safe to do

```powershell
if ($Value -is [System.Collections.IList]) {
    foreach ($e in $Value) { ConvertTo-PsNative -Value $e }
}
```

plugins/modules/win_pssession_configuration.ps1
```diff
@@ -203,6 +203,155 @@ function ConvertFrom-AnsibleOption {
...
+    )
+
+    End {
+        if ($null -eq $Value) { return $null }
+        # Port of C# ParseList: perform .NET type checks to mirror behavior exactly.
+        $valueType = $Value.GetType()
+
+        # We'll collect parsed elements first, then coerce each element into the
+        # final shape (string or Hashtable) expected by New-PSSessionConfigurationFile.
+        $out = @()
```
jborean93: It's not great to do `+=` on an array. While the sizes we are dealing with won't really matter we can just do

```powershell
$out = @(
    if ($Value -is ...) {
         ... emit objects like normal
    }
    elseif ($Value ...) {
        ... More emitting
    }
)
```

PowerShell will collect all the output and coerce into an array all for you.

plugins/modules/win_pssession_configuration.ps1
```diff
@@ -203,6 +203,155 @@ function ConvertFrom-AnsibleOption {
...
+        }
+        else {
+            throw "{0} cannot be converted to a list" -f $valueType.FullName
+        }
+
+        # Now coerce each parsed element into either a Hashtable (for dict-like
+        # values) or a string. Also flatten nested enumerables that may have been
+        # produced by ConvertTo-PsNative.
+        $final = @()
+        foreach ($e in $out) {
```
jborean93: Is there a reason that `ConvertTo-PsNative` can't handle all this logic, it seems to already? Seems wrong to do a second pass through on the values.

plugins/modules/win_pssession_configuration.ps1
```diff
@@ -203,6 +203,155 @@ function ConvertFrom-AnsibleOption {
...
+            foreach ($k in $Value.Keys) {
+                $v = $Value[$k]
+                if ($k -is [string]) { $keyName = $k } else { $keyName = $k.ToString() }
+                $ht[$keyName] = ConvertTo-PsNative -Value $v
+            }
+            return $ht
+        }
+
+        # IEnumerable (but not string) -> array of parsed elements
+        if ($Value -is [System.Collections.IEnumerable] -and -not ($Value -is [string])) {
```
jborean93: I think we can just have this check for `IList` to remove the `String` check and make it simpler. I cannot think of any reason why we wouldn't have anything but an `IList` here

```powershell
if ($Value -is [System.Collections.IList]) {
}
```

plugins/modules/win_pssession_configuration.ps1
```diff
@@ -203,6 +203,155 @@ function ConvertFrom-AnsibleOption {
...
+            $out = @()
+            foreach ($e in $Value) { $out += (ConvertTo-PsNative -Value $e) }
+            return $out
```
jborean93: No need to collect into another array, just output it

```suggestion
            foreach ($e in $Value) { ConvertTo-PsNative -Value $e }
```

PR #682, opened 2026-04-21
win_unzip: use native tar.exe where available

noamper:
Hoping to avoid the PSCX dependency for Server 2019+ where native tar.exe extraction meets our needs. These changes are LLM-generated. I have reviewed and guided them, but lack much experience with Ansible.  

- Uses `%SystemRoot%\System32\tar.exe` (ships with Windows 10 build 17063+ and Server 2019+) to extract tar-based archives (`.tar`,`.tar.gz`/`.tgz`, `.tar.bz2`/`.tbz2`, `.tar.xz`/`.txz`) without requiring PSCX. 
- Falls back to PSCX on older systems where `tar.exe` is not present. 
- Fixes a pre-existing bug where `Expand-Archive -Path` wildcard-expanded paths containing square brackets and other special characters; changed to `-LiteralPath` in both call sites.  

Integration tests extended with 

- tar.gz check mode, extraction (content verified), delete-in-check-mode, and delete-actual 
- PSCX regression test, skipped when PSCX is not installed

Tested locally against Windows Server 2019 with & without PSCX installed. The tar tests run independently of PSCX.

Changelog  `changelogs/fragments/win_unzip-tar-exe.yml` included.

### DIFF
--- a/changelogs/fragments/win_unzip-tar-exe.yml
+++ b/changelogs/fragments/win_unzip-tar-exe.yml
@@ -1,0 +1,7 @@
+minor_changes:
+  - >-
+    win_unzip - Use ``tar.exe`` from ``%SystemRoot%\System32`` on Windows 10 build 17063 and later
+    and Windows Server 2019 and later to extract tar-based archives (``.tar``, ``.tar.gz``/``.tgz``,
+    ``.tar.bz2``/``.tbz2``, ``.tar.xz``/``.txz``) without requiring PSCX. On older systems where
+    ``tar.exe`` is not available, PSCX remains the fallback. PSCX is still required for passwords,
+    recursive extraction, and non-tar formats other than ``.zip``.
--- a/changelogs/fragments/win_unzip-tar-exe.yml
+++ b/changelogs/fragments/win_unzip-tar-exe.yml
@@ -5,3 +5,8 @@ minor_changes:
     ``.tar.bz2``/``.tbz2``, ``.tar.xz``/``.txz``) without requiring PSCX. On older systems where
     ``tar.exe`` is not available, PSCX remains the fallback. PSCX is still required for passwords,
     recursive extraction, and non-tar formats other than ``.zip``.
+bugfixes:
+  - >-
+    win_unzip - Use ``-LiteralPath`` when calling PSCX ``Expand-Archive`` so paths containing
+    PowerShell wildcard characters (e.g. ``[``, ``]``) are not glob-expanded and silently
+    resolved to an empty array.
--- a/plugins/modules/win_unzip.ps1
+++ b/plugins/modules/win_unzip.ps1
@@ -203,7 +203,8 @@ If ($use_pscx) {
     $list = Get-Module -ListAvailable
 
     If (-Not ($list -match "PSCX")) {
-        Fail-Json -obj $result -message "PowerShellCommunityExtensions PowerShell Module (PSCX) is required for this archive type, recursive extraction, or password-protected archives."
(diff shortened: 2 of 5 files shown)


plugins/modules/win_unzip.ps1
```diff
@@ -84,6 +84,40 @@ Function Expand-ZipLegacy($src, $dest) {
...
+    try {
+        $tar_output = & $tar_exe -xf $src -C $dest 2>&1
+        $failed = $LASTEXITCODE -ne 0
+    }
+    catch {
+        $failed = $true
+        $tar_output = $_.Exception.Message
+    }
+    finally {
+        $ErrorActionPreference = $prev_eap
```
jborean93: No need to set this back, `$ErrorActionPreference` gets scoped to where it is set so you can just do `$ErrorActionPreference = 'Continue'` before calling the exe and avoid the finally block. We also have a module util you could use to capture the streams separately but that's overkill for the purposes here.

plugins/modules/win_unzip.ps1
```diff
@@ -84,6 +84,40 @@ Function Expand-ZipLegacy($src, $dest) {
...
+    }
+    if ($failed) {
+        Fail-Json -obj $result -message "Error extracting '$src' to '$dest' using tar.exe: $tar_output"
```
jborean93: It might be a good idea to include the `$LASTEXITCODE` and use that as an indicator. e.g. set `$rc = 0` before try, set `$rc = $LASTEXITCODE` after calling, `$rc = -1` in the catch block and have `if ($rc)` for this if condition. We can then include rc in the message.

plugins/modules/win_unzip.py
```diff
@@ -10,11 +10,18 @@
 short_description: Unzips compressed files and archives on the Windows node
 description:
 - Unzips compressed files and archives.
-- Supports .zip files natively.
-- Supports other formats supported by the Powershell Community Extensions (PSCX) module (basically everything 7zip supports).
+- Supports .zip files natively via the .NET BCL.
+- Supports tar-based formats (C(.tar), C(.tar.gz)/C(.tgz), C(.tar.bz2)/C(.tbz2), C(.tar.xz)/C(.txz))
```
jborean93, line 14: I would try and include that this support was added in `community.windows 3.2.0` (the next version of this collection).

changelogs/fragments/win_unzip-tar-exe.yml
```diff
@@ -0,0 +1,7 @@
+minor_changes:
```
jborean93, line 1: You probably also want a `bugfixes` entry for the `-LiteralPath` fix you added as well.

jborean93 (changes requested):
Thanks for working on this. Only have some very minor review comments but the rest looks fantastic. Thanks so much for adding the integration tests here.

changelogs/fragments/win_unzip-tar-exe.yml
noamper: Thanks, added

plugins/modules/win_unzip.ps1
noamper: Interesting, good to know. Dropped the finally.

plugins/modules/win_unzip.ps1
noamper: Yes! Added.

jborean93 (approved):

jborean93:
Thank you very much for working on this, changes look great!

PR #687, opened 2026-06-06
fix: win_scheduled_task - Error creating task when ansible user is SYSTEM

### LINKED ISSUE #633: win_scheduled_task failed to register new task definition: (32,4):UserId:
##### SUMMARY
I'm trying to convert a `win_shell` task to `win_scheduled_task` and it's failing with `failed to register new task definition: (32,4):UserId:`. I tried updating the collection to 3.0.0 but behavior is the same. The error message appears to be from _plugins/modules/win_scheduled_task.ps1#L1107_. I can't seem to find MS docs that explain what the error means. Looking at the win_scheduled_task.ps1 and the Win32 docs for  [TaskFolder.RegisterTaskDefinition](https://learn.microsoft.com/en-us/windows/win32/taskschd/taskfolder-registertaskdefinition) I would guess that the task definition is missing the UserId. Though I'm more inclined to believe the error is in my implementation.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
community.windows.win_scheduled_task

##### ANSIBLE VERSION

```
$ ansible --version
ansible [core 2.16.14]
  config file = /code/packer/windows2022-baseimage-pipeline/ansible.cfg
  configured module search path = ['/home/justin/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.12/dist-packages/ansible
  ansible collection location = /home/justin/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/bin/ansible
  python version = 3.12.3 (main, Feb  4 2025, 14:48:35) [GCC 13.3.0] (/usr/bin/python3)
  jinja version = 3.1.6
  libyaml = True
```

##### COLLECTION VERSION

```
# /usr/local/lib/python3.12/dist-packages/ansible_collections
Collection        Version
----------------- -------
community.windows 2.3.0  
```

##### CONFIGURATION

```
$ ansible-config dump --only-changed
ANSIBLE_FORCE_COLOR(env: ANSIBLE_FORCE_COLOR) = True
CONFIG_FILE() = /code/packer/windows2022-baseimage-pipeline/ansible.cfg
$ cat /code/packer/windows2022-baseimage-pipeline/ansible.cfg 
[defaults]
callback_whitelist = time, profile_tasks
```

##### OS / ENVIRONMENT

Target machine is AWS instance from Amazon's Windows_Server-2022-English-Full-Base image

##### STEPS TO REPRODUCE

```yaml
- name: Reprovisioning WinRM Certificate
  become: yes
  become_flags: logon_type=batch
  become_method: runas
  become_user: SYSTEM
  block:
    - name: Create mgmttask directory
      ansible.windows.win_file:
        path: C:\mgmttasks
        state: directory

    - name: Copy winrm_fix.ps1 in destination image
      ansible.windows.win_copy:
        src: files/winrm_fix.ps1
        dest: C:\mgmttasks\winrm_fix.ps1
      register: winrm_fix_copy

    # - name: Create Scheduled task to run-once
    #   ansible.windows.win_shell: |
    #     $destination = '-WindowStyle hidden {{ winrm_fix_copy.dest }} -ForceNewSSLCert'
    #     $action = New-ScheduledTaskAction -Execute 'powershell.exe' -Argument $destination
    #     $trigger = New-ScheduledTaskTrigger -AtStartup
    #     $principal = New-ScheduledTaskPrincipal -UserID "NT AUTHORITY\SYSTEM" -LogonType ServiceAccount -RunLevel Highest
    #     Register-ScheduledTask -TaskName 'WinRMReprovisioning' -TaskPath 'Hotfixes' -Action $action \
    #       -Trigger $trigger -Principal $principal -Description 'WinRM Reprovisioning certificate' -ErrorAction Stop
    
    - name: Create run-once scheduled WinRMReprovisioning
      community.windows.win_scheduled_task:
        name: WinRMReprovisioning
        display_name: Reconfigure WinRM on first boot
        actions:
          - arguments: -WindowStyle hidden {{ winrm_fix_copy.dest }} -ForceNewSSLCert
            path: powershell.exe
        enabled: true
        logon_type: service_account
        path: \Hotfixes
        run_level: highest
        state: present
        triggers:
          - type: boot
        username: SYSTEM
```

##### EXPECTED RESULTS

Scheduled task is created on target machine

##### ACTUAL RESULTS

```
TASK [security-agents : Create mgmttask directory] ***************************************************************
task path: /home/justin/.ansible/roles/security-agents/tasks/windows-scheduler.yml:62
ok: [ip-10-127-136-10.ec2.internal] => {"changed": false}

TASK [security-agents : Copy winrm_fix.ps1 in destination image] *************************************************
task path: /home/justin/.ansible/roles/security-agents/tasks/windows-scheduler.yml:67
ok: [ip-10-127-136-10.ec2.internal] => {"changed": false, "checksum": "cffe902cb83d8ad0ffe0c16888167e649ffd767e", "dest": "C:\\mgmttasks\\winrm_fix.ps1", "operation": "file_copy", "original_basename": "winrm_fix.ps1", "size": 15999, "src": "files/winrm_fix.ps1"}

TASK [security-agents : Create run-once scheduled WinRMReprovisioning] *******************************************
task path: /home/justin/.ansible/roles/security-agents/tasks/windows-scheduler.yml:82
fatal: [ip-10-127-136-10.ec2.internal]: FAILED! => {"changed": false, "msg": "failed to register new task definition: (32,4):UserId:"}
```

stanleyz:
##### SUMMARY

The TaskFolder::RegisterTaskDefinition method returns error when called by the SYSTEM account with the user parameter equal to NULL, this PR fixes it by setting the user parameter to the username supplied or otherwise if empty to the current username.

- win_scheduled_task - Error creating scheduled tasks with SYSTEM account. Fixed by setting the user parameter to the username supplied or otherwise if empty to the current username. (https://github.com/MicrosoftDocs/sdk-api/blob/docs/sdk-api-src/content/taskschd/nf-taskschd-itaskfolder-registertaskdefinition.md)

This fixes #633 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

win_scheduled_taksk

##### ADDITIONAL INFORMATION

Before:
```
TASK [win_scheduled_task : task with password principal] ***********************
[ERROR]: Task failed: Module failed: failed to register new task definition: (31,4):UserId:
Origin: /home/stanley/.ansible/collections/ansible_collections/community/windows/tests/output/.tmp/integration/win_scheduled_task-iokru30h-ÅÑŚÌβŁÈ/tests/integration/targets/win_scheduled_task/tasks/principals.yml:35:3

33     - task_with_password_result_check.task_exists == False
34
35 - name: task with password principal
     ^ column 3

fatal: [WindowsServer]: FAILED! => {"changed": false, "msg": "failed to register new task definition: (31,4):UserId:"}

```
After
```
TASK [win_scheduled_task : task with password principal] ***********************
changed: [WindowsServer]                                                               

TASK [win_scheduled_task : get result of task with password principal] *********
ok: [WindowsServer]
                                                                                                           
TASK [win_scheduled_task : assert results of task with password principal] *****
ok: [WindowsServer] => {             
    "changed": false,                                     
    "msg": "All assertions passed"
}
```

Reference doc:

https://github.com/MicrosoftDocs/sdk-api/blob/docs/sdk-api-src/content/taskschd/nf-taskschd-itaskfolder-registertaskdefinition.md
https://learn.microsoft.com/en-us/windows/win32/api/taskschd/nf-taskschd-itaskfolder-registertaskdefinition

### DIFF
--- a/changelogs/fragments/scheduled_task-system.yml
+++ b/changelogs/fragments/scheduled_task-system.yml
@@ -1,0 +1,4 @@
+bugfixes:
+  - >-
+    win_scheduled_task - Fix issue when creating a new scheduled task when using ``become_user: SYSTEM`` -
+    https://github.com/ansible-collections/community.windows/issues/633
--- a/plugins/modules/win_scheduled_task.ps1
+++ b/plugins/modules/win_scheduled_task.ps1
@@ -1103,8 +1103,12 @@ else {
             }
         }
 
+        $tmp_user = $username
+        if ($null -eq $tmp_user) {
+            $tmp_user = [System.Security.Principal.WindowsIdentity]::GetCurrent().Name
+        }
         try {
-            $task = $task_folder.RegisterTaskDefinition($name, $task_definition, $task_creation_flags, $null, $null, $null)
+            $task = $task_folder.RegisterTaskDefinition($name, $task_definition, $task_creation_flags, $tmp_user, $null, $null)
         }
         catch {
             Fail-Json -obj $result -message "failed to register new task definition: $($_.Exception.Message)"
--- a/tests/integration/targets/win_scheduled_task/tasks/principals.yml
+++ b/tests/integration/targets/win_scheduled_task/tasks/principals.yml
@@ -43,6 +43,9 @@
     actions:
     - path: cmd.exe
   register: task_with_password
+  become: true
+  become_user: SYSTEM
+  become_method: runas
 
 - name: get result of task with password principal
   win_scheduled_task_stat:
--- a/tests/integration/targets/win_scheduled_task/tasks/principals.yml
+++ b/tests/integration/targets/win_scheduled_task/tasks/principals.yml
@@ -43,9 +43,6 @@
     actions:
     - path: cmd.exe
   register: task_with_password
-  become: true
-  become_user: SYSTEM
-  become_method: runas
 
 - name: get result of task with password principal
   win_scheduled_task_stat:
--- a/tests/integration/targets/win_scheduled_task/tasks/tests.yml
+++ b/tests/integration/targets/win_scheduled_task/tasks/tests.yml
@@ -439,12 +439,14 @@
     that:
     - remove_sole_task_again is not changed
 
-- name: with SYSTEM user
+# Task Scheduler seems to have some special requirements when running as SYSTEM
+# https://github.com/ansible-collections/community.windows/issues/633
+- name: with SYSTEM user tests
   become: true
   become_user: SYSTEM
   become_method: runas
   block:
-  - name: create task
+  - name: create task as SYSTEM user
     win_scheduled_task:
       name: "{{test_scheduled_task_name}} with system user"
       path: "{{test_scheduled_task_path}}"
@@ -455,13 +457,13 @@
       description: Original Description
     register: create_task
 
-  - name: get result of create task 
+  - name: get result of create task as SYSTEM user
     win_scheduled_task_stat:
       path: "{{test_scheduled_task_path}}"
       name: "{{test_scheduled_task_name}} with system user"
     register: create_task_result
 
-  - name: Remove task
+  - name: Remove task as SYSTEM user
     win_scheduled_task:
       name: "{{test_scheduled_task_name}} with system user"
       path: "{{test_scheduled_task_path}}"
--- a/tests/integration/targets/win_scheduled_task/tasks/tests.yml
+++ b/tests/integration/targets/win_scheduled_task/tasks/tests.yml
@@ -438,3 +438,31 @@
   assert:
     that:
     - remove_sole_task_again is not changed
+
+- name: with SYSTEM user
+  become: true
+  become_user: SYSTEM
+  become_method: runas
+  block:
+  - name: create task
+    win_scheduled_task:
+      name: "{{test_scheduled_task_name}} with system user"
+      path: "{{test_scheduled_task_path}}"
+      state: present
+      actions:
+      - path: cmd.exe
+        arguments: /c echo hi
+      description: Original Description
+    register: create_task
+
+  - name: get result of create task 
+    win_scheduled_task_stat:
+      path: "{{test_scheduled_task_path}}"
+      name: "{{test_scheduled_task_name}} with system user"
+    register: create_task_result
+
+  - name: Remove task
+    win_scheduled_task:
+      name: "{{test_scheduled_task_name}} with system user"
+      path: "{{test_scheduled_task_path}}"
+      state: absent


jborean93:
Weirdly I cannot replicate this locally on my dev machine but testing it in CI with become SYSTEM does fail https://github.com/ansible-collections/community.windows/pull/690. I'll have to try and do some research to figure out what exactly might be happening here and if the suggested change is what we do want.

Before we can merge this change though we would need a changelog fragment https://docs.ansible.com/projects/ansible/latest/community/development_process.html#changelog-fragments to document this bugfix.

I also think the test you've changed may need some tweaking. While it certainly is testing this scenario it's not exactly clear why we are running it as SYSTEM. It would be better to have an explicit test with a `name: ` that describes exactly what we are testing. Maybe we can move that into one of the `tests.yml` tasks instead?

stanleyz:
> Before we can merge this change though we would need a changelog fragment https://docs.ansible.com/projects/ansible/latest/community/development_process.html#changelog-fragments to document this bugfix.

How about the below look like, I've include it in the above description as well.

* win_scheduled_task - Error creating scheduled tasks with SYSTEM account. Fixed by setting the user parameter to the username supplied or otherwise if empty to the current username. (https://github.com/MicrosoftDocs/sdk-api/blob/docs/sdk-api-src/content/taskschd/nf-taskschd-itaskfolder-registertaskdefinition.md)

> I also think the test you've changed may need some tweaking. While it certainly is testing this scenario it's not exactly clear why we are running it as SYSTEM. It would be better to have an explicit test with a `name: ` that describes exactly what we are testing. Maybe we can move that into one of the `tests.yml` tasks instead?

Yeah, I unerstand. Because the error only happens when a new scheduled tasks is created, to test it, we'd need a separate file (tests) just for this. I actually started that way but feel a bit overkill, hence revert back to what it is now. Happy to swtich it back to that way if that's what you want, or otherwise I can reword the existing tasks a bit or even with a comment to make it clearer?

jborean93:
> How about the below look like, I've include it in the above description as well.

It is fine but I would probably recommend a more human/user friendly message like, fixed issue when creating a task when running as the `SYSTEM` user. If you think your message is better then that's totally fine, I usually just try and craft the entries to reflect how the bugfix affects a user than what the actual fix was.

> Happy to swtich it back to that way if that's what you want

I think so, the tests are pretty verbose right now but they are designed to make sure we don't regress with these edge cases. I think adding a new test at the end of https://github.com/ansible-collections/community.windows/blob/main/tests/integration/targets/win_scheduled_task/tasks/tests.yml that adds and removes a simple task when running as `SYSTEM` should suffic for this use case.

stanleyz:
> It is fine but I would probably recommend a more human/user friendly message like, fixed issue when creating a task when running as the `SYSTEM` user. If you think your message is better then that's totally fine, I usually just try and craft the entries to reflect how the bugfix affects a user than what the actual fix was.

cool, I am too fussy about this, leave that wording to you then.

> I think so, the tests are pretty verbose right now but they are designed to make sure we don't regress with these edge cases. I think adding a new test at the end of https://github.com/ansible-collections/community.windows/blob/main/tests/integration/targets/win_scheduled_task/tasks/tests.yml that adds and removes a simple task when running as `SYSTEM` should suffic for this use case.

Added, please check, thanks.

jborean93 (approved):

jborean93:
Thanks for working on this, just pushed a changelog fragment and a slight tweak to the test names.

PR #688, opened 2026-06-15
Improve performance of win_psmodule and win_psmodule_info modules

### LINKED ISSUE #649: win_psmodule: task takes forever if host has many PS modules
##### SUMMARY
When executing the `win_psmodule` task on a remote Windows host that has a large number of Powershell modules (or even a few large Powershell modules), the task has a long execution time even if the task is not installing the Powershell module.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
win_psmodule

##### ANSIBLE VERSION
```
ansible [core 2.18.2rc1.post0]
  config file = /home/shawn.hardwick/ansible.cfg
  configured module search path = ['/home/shawn.hardwick/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/shawn.hardwick/code/venv/ansible-2.18/lib/python3.12/site-packages/ansible
  ansible collection location = /home/shawn.hardwick/code/ansible-projects/collections
  executable location = /home/shawn.hardwick/code/venv/ansible-2.18/bin/ansible
  python version = 3.12.9 (main, Feb  5 2025, 08:49:00) [GCC 11.4.0] (/home/shawn.hardwick/code/venv/ansible-2.18/bin/python3.12)
  jinja version = 3.1.5
  libyaml = True
```

##### COLLECTION VERSION
``
Collection        Version
----------------- -------
community.windows 2.4.0  
```

##### CONFIGURATION
```
N/A
```

##### OS / ENVIRONMENT
Controller: Ubuntu Focal 24.04 on WSL2
Target: Windows Server 2022


##### STEPS TO REPRODUCE
Target a Windows remote host with many (100+) Powershell modules (like AWS.Tools.*) or a few large Powershell modules (like AWSPowershell) in addition to the PS module being installed by Ansible.

<!--- Paste example playbooks or commands between quotes below -->
```yaml
---
- name: Install Powershell Module
  hosts: all
  tasks:
    - name: "Install Powershell Module - xRemoteDesktopAdmin"
      community.windows.win_psmodule:
        name: "xRemoteDesktopAdmin"
        state: Present
```
Observe large runtimes by the task even if the task does not install the Powershell module.

##### EXPECTED RESULTS
Takes takes a few seconds if no installation is needed

##### ACTUAL RESULTS
Task takes 35 seconds
```
[2025-08-27T11:46:29] TASK [rdp : Install Powershell Module - xRemoteDesktopAdmin] ***
[2025-08-27T11:47:04] ok: [server.example.com]
```

##### Additional Notes

The main issue stems from the usage of `Get-Module -ListAvailable`  within `win_psmodule` which requires Powershell to parse the module metadata and all commands for every Powershell Module on the filesystem.
https://github.com/ansible-collections/community.windows/blob/main/plugins/modules/win_psmodule.ps1#L159

This means that increased parsing time affects installation time too, but it is less evident/obvious.

This may also affect `win_psmodule_info`, but I haven't tested it.

A more efficient method would be using `Get-Module -ListAvailable $Name` which would only parse modules matching the criteria. I'm unsure of the backwards compatibility of this approach for older Powershell versions.

For our use case, we have the `AWSPowershell` Powershell module installed (which comes by default on AWS AMIs) which can take 10-15 seconds to parse on top of `AWS.Tools` which can add an additional 10 seconds.
```
# On a system with both AWSPowershell and AWS.Tools installed
PS C:\Users\user> (Measure-Command { Get-Module -ListAvailable }).TotalSeconds
31.2059538
PS C:\Users\user> (Measure-Command { Get-Module -ListAvailable AWSPowershell }).TotalSeconds
14.8539835
PS C:\Users\user> (Measure-Command { Get-Module -ListAvailable AWS.Tools.* }).TotalSeconds
9.8902874
PS C:\Users\user> (Measure-Command { Get-Module -ListAvailable xRemoteDesktopAdmin }).TotalSeconds
0.2322632
PS C:\Users\user> $(Get-Module -ListAvailable | Measure-Object).Count
511
```

ShawnHardwick:
**SUMMARY**
- Add `skip_module_repository_info` and `include_properties` support to win_psmodule_info for faster, more customizable module metadata retrieval.
- Optimize `win_psmodule` by querying `Get-Module -ListAvailable -Name $Name` instead of enumerating all modules, reducing overhead on systems with many installed modules.

**COMPONENT NAME**
`win_psmodule_info`
`win_psmodule`

**ADDITIONAL INFORMATION**

Fixes https://github.com/ansible-collections/community.windows/issues/649

For remote Windows hosts that have many PS modules installed (like `AWS.Tools` ), `win_psmodule` would take ~20 seconds to  evaluate all of the modules on the host when we only needed to retrieve the one being installed.

```
# Current module (~23 seconds
[2026-06-12 15:09:05] TASK [collection.example.rdp : Install Powershell Module - xRemoteDesktopAdmin] ***
[2026-06-12 15:09:28] ok: [windows-server-2022-full-base]

# New module (~4 seconds)
[2026-06-12 15:03:07] TASK [collection.example.rdp : Install Powershell Module - xRemoteDesktopAdmin] ***
[2026-06-12 15:03:11] ok: [windows-server-2022-full-base]
```

Additionally, for Windows hosts with many PS module installed, the `win_psmodule_info` would return way too much data, especially when modules like AWS.Tools have hundreds of exported modules and aliases. Sending this data over WinRM was expensive when a user may only be interested in a few properties. This is even more important when running Ansible with `ansible-runner` which will json-ify the entire output of the task and fill the terminal/log with lots of unnecessary data.

### DIFF
--- a/changelogs/fragments/win_psmodule_info_include_properties.yml
+++ b/changelogs/fragments/win_psmodule_info_include_properties.yml
@@ -1,21 +1,21 @@
 ---
 minor_changes:
   - >-
-    community.windows.win_psmodule_info - 
-    Added ``include_properties`` parameter to allow fine-grained control over which module properties are returned, 
-    improving performance when only specific properties are needed 
+    community.windows.win_psmodule_info -
+    Added ``include_properties`` parameter to allow fine-grained control over which module properties are returned,
+    improving performance when only specific properties are needed
     (https://github.com/ansible-collections/community.windows/pull/688).
   - >-
-    community.windows.win_psmodule_info - 
-    Added ``skip_module_repository_info`` parameter to skips querying PowerShellGet for repository-related metadata 
+    community.windows.win_psmodule_info -
+    Added ``skip_module_repository_info`` parameter to skips querying PowerShellGet for repository-related metadata
     (https://github.com/ansible-collections/community.windows/pull/688).
   - >-
-    community.windows.win_psmodule_info - 
-    Automatically skips expensive PowerShellGet repository lookups when ``include_properties`` is specified without repository-related properties 
+    community.windows.win_psmodule_info -
+    Automatically skips expensive PowerShellGet repository lookups when ``include_properties`` is specified without repository-related properties
     (https://github.com/ansible-collections/community.windows/pull/688).
 
 bugfixes:
   - >-
-    community.windows.win_psmodule_info - 
-    Fixed typo in documentation changing ``procoessor_architecture`` to ``processor_architecture`` 
+    community.windows.win_psmodule_info -
+    Fixed typo in documentation changing ``procoessor_architecture`` to ``processor_architecture``
     (https://github.com/ansible-collections/community.windows/pull/688).
--- a/changelogs/fragments/win_psmodule_info_include_properties.yml
+++ b/changelogs/fragments/win_psmodule_info_include_properties.yml
@@ -1,8 +1,21 @@
 ---
 minor_changes:
-  - flatiron.windows.win_psmodule_info - Added ``include_properties`` parameter to allow fine-grained control over which module properties are returned, improving performance when only specific properties are needed (https://github.com/ansible-collections/community.windows/pull/688).
-  - flatiron.windows.win_psmodule_info - Added ``skip_module_repository_info`` parameter to skips querying PowerShellGet for repository-related metadata (https://github.com/ansible-collections/community.windows/pull/688).
-  - flatiron.windows.win_psmodule_info - Automatically skips expensive PowerShellGet repository lookups when ``include_properties`` is specified without repository-related properties (https://github.com/ansible-collections/community.windows/pull/688).
+  - >-
+    community.windows.win_psmodule_info - 
+    Added ``include_properties`` parameter to allow fine-grained control over which module properties are returned, 
+    improving performance when only specific properties are needed 
+    (https://github.com/ansible-collections/community.windows/pull/688).
+  - >-
+    community.windows.win_psmodule_info - 
+    Added ``skip_module_repository_info`` parameter to skips querying PowerShellGet for repository-related metadata 
+    (https://github.com/ansible-collections/community.windows/pull/688).
+  - >-
+    community.windows.win_psmodule_info - 
+    Automatically skips expensive PowerShellGet repository lookups when ``include_properties`` is specified without repository-related properties 
+    (https://github.com/ansible-collections/community.windows/pull/688).
 
 bugfixes:
-  - flatiron.windows.win_psmodule_info - Fixed typo in documentation changing ``procoessor_architecture`` to ``processor_architecture`` (https://github.com/ansible-collections/community.windows/pull/688).
+  - >-
+    community.windows.win_psmodule_info - 
+    Fixed typo in documentation changing ``procoessor_architecture`` to ``processor_architecture`` 
+    (https://github.com/ansible-collections/community.windows/pull/688).
--- a/changelogs/fragments/win_psmodule_info_include_properties.yml
+++ b/changelogs/fragments/win_psmodule_info_include_properties.yml
@@ -1,8 +1,8 @@
 ---
 minor_changes:
-  - flatiron.windows.win_psmodule_info - Added C(include_properties) parameter to allow fine-grained control over which module properties are returned, improving performance when only specific properties are needed.
-  - flatiron.windows.win_psmodule_info - Added C(skip_module_repository_info) parameter to skips querying PowerShellGet for repository-related metadata.
-  - flatiron.windows.win_psmodule_info - Automatically skips expensive PowerShellGet repository lookups when C(include_properties) is specified without repository-related properties.
+  - flatiron.windows.win_psmodule_info - Added ``include_properties`` parameter to allow fine-grained control over which module properties are returned, improving performance when only specific properties are needed (https://github.com/ansible-collections/community.windows/pull/688).
+  - flatiron.windows.win_psmodule_info - Added ``skip_module_repository_info`` parameter to skips querying PowerShellGet for repository-related metadata (https://github.com/ansible-collections/community.windows/pull/688).
+  - flatiron.windows.win_psmodule_info - Automatically skips expensive PowerShellGet repository lookups when ``include_properties`` is specified without repository-related properties (https://github.com/ansible-collections/community.windows/pull/688).
 
 bugfixes:
-  - flatiron.windows.win_psmodule_info - Fixed typo in documentation changing C(procoessor_architecture) to C(processor_architecture) (MR-URL).
+  - flatiron.windows.win_psmodule_info - Fixed typo in documentation changing ``procoessor_architecture`` to ``processor_architecture`` (https://github.com/ansible-collections/community.windows/pull/688).
--- a/changelogs/fragments/win_psmodule_info_include_properties.yml
+++ b/changelogs/fragments/win_psmodule_info_include_properties.yml
@@ -1,0 +1,8 @@
+---
+minor_changes:
+  - flatiron.windows.win_psmodule_info - Added C(include_properties) parameter to allow fine-grained control over which module properties are returned, improving performance when only specific properties are needed.
+  - flatiron.windows.win_psmodule_info - Added C(skip_module_repository_info) parameter to skips querying PowerShellGet for repository-related metadata.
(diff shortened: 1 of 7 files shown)


changelogs/fragments/win_psmodule_info_include_properties.yml
```diff
@@ -0,0 +1,8 @@
+---
+minor_changes:
+  - flatiron.windows.win_psmodule_info - Added C(include_properties) parameter to allow fine-grained control over which module properties are returned, improving performance when only specific properties are needed.
```
briantist: the `C()` formatting doesn't work here, need to use double backticks like 
```
...Added ``include_properties`` ...
```

briantist: Also all lines should have a URL at the end (preferably the issue, otherwise the PR link)

plugins/modules/win_psmodule.ps1
```diff
@@ -147,7 +147,7 @@ Function Get-PsModule {
         Version = ""
     }
 
-    $ExistingModules = Get-Module -ListAvailable | Where-Object { ($_.name -eq $Name) }
+    $ExistingModules = Get-Module -ListAvailable -Name $Name
```
briantist: These are subtly different. This module does not support wildcards (since it's for installing or remove a specific module). The old version of this command will not return any results with a wildcard (the condition uses `-eq`) but your version will, which may lead to unpredictable behavior.

I can't tell yet whether that would cause any actual issues or not.

A possible simple fix is to use the `-Name` parameter and then keep the `Where-Object` call to further filter out wildcard results. But I'm not sure if that's foolproof either just an idea.

plugins/modules/win_psmodule.ps1
```diff
@@ -293,7 +293,7 @@ Function Remove-PsModule {
         [Bool]$CheckMode
     )
     # If module is present, uninstalls it.
-    if (Get-Module -ListAvailable | Where-Object { $_.name -eq $Name }) {
+    if (Get-Module -ListAvailable -Name $Name) {
```
briantist: See above comment

plugins/modules/win_psmodule_info.ps1
```diff
@@ -11,11 +11,34 @@ $spec = @{
...
+    'PackageManagementProvider',
+    'InstalledLocation',
+    'RepositorySourceLocation',
+    'Tags',
+    'CompatiblePSEditions',
+    'LicenseUri',
+    'ProjectUri',
+    'IconUri',
+    'ReleaseNotes',
+    'ExportedDscResources',
```
briantist: We can actually remove the trailing commas ',' in a multi-line list in PowerShell, something I forgot to do originally 🤦 but since we're modifying it now anyway might as well

plugins/modules/win_psmodule_info.ps1
```diff
@@ -236,24 +292,8 @@ function Add-ModuleRepositoryInfo {
...
-            'RepositorySourceLocation',
-            'Tags',
-            'CompatiblePSEditions',
-            'LicenseUri',
-            'ProjectUri',
-            'IconUri',
-            'ReleaseNotes',
-            'ExportedDscResources', # ExportedDscResources is not returned here, this is a hack for Windows 2012/R2 to ensure the field is present
-            'Prefix'                # Prefix is not actually returned here, this is a hack for Windows 2012 just to ensure the field is present
-        )
```
briantist, line 256: Let's ensure these comments get moved to this list's new location

changelogs/fragments/win_psmodule_info_include_properties.yml
ShawnHardwick: Ah yes, that reminds me to also reference my previously submitted issue in my MR

plugins/modules/win_psmodule.ps1
ShawnHardwick: That's an interesting edge case. 

So if a user gave something like `AWS.Tools.*` as a `Name`, it may or may not find a matching module. In case it does find a module, it would give success when in truth it should probably fail. If no module is found, it would attempt to install `AWS.Tools.*`, and I don't know if wildcards are accepted for installing module.

I don't think wildcards should be an accepted criteria, right?

Your suggestion to include both filters seems to handle this

plugins/modules/win_psmodule_info.ps1
ShawnHardwick: Done

plugins/modules/win_psmodule_info.ps1
ShawnHardwick: Done

plugins/modules/win_psmodule.ps1
briantist: Right, I don't think wildcards are acceptable for `win_psmodule` since the underlying APIs (for installation/removal) don't support them, but yeah I think both filters would keep the current behavior

ShawnHardwick:
Think that should address all of the comments and a few linting issues I saw. Not sure whats going on with the pipelines though.

changelogs/fragments/win_psmodule_perf.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
```
jborean93: Minor nit, I feel like this is more on the bugfixes side than a feature

changelogs/fragments/win_psmodule_info_include_properties.yml
```diff
@@ -0,0 +1,8 @@
+---
+minor_changes:
+  - flatiron.windows.win_psmodule_info - Added ``include_properties`` parameter to allow fine-grained control over which module properties are returned, improving performance when only specific properties are needed (https://github.com/ansible-collections/community.windows/pull/688).
```
jborean93: This an all the other changelog entries should be `community.windows.*` and not `flatiron.windows.*`.

plugins/modules/win_psmodule_info.py
```diff
@@ -26,6 +26,28 @@
...
+      - List of property names to include in the output.
+      - Property names should be in snake_case format (e.g., C(module_base), C(exported_functions)).
+      - If omitted, all available properties are returned.
+      - When specified, only the requested properties will be included in the output.
+      - To get minimal output similar to the old C(simplified=true), use C([name, version, author, module_base]).
+      - See the RETURN section for a complete list of available properties.
+      - Repository-related properties (repository, repository_source_location, installed_date, published_date, etc.)
+        require PowerShellGet lookups and may be slower.
+    type: list
+    elements: str
```
jborean93, line 40: A new option should have `version_added` which will be `3.3.0` for the new feature release.

```suggestion
    elements: str
    version_added: 3.3.0
```

plugins/modules/win_psmodule_info.py
```diff
@@ -26,6 +26,28 @@
...
+  skip_module_repository_info:
+    description:
+      - When C(true), skips querying PowerShellGet for repository-related metadata.
+      - This significantly improves performance but repository properties will be C(null).
+      - Repository properties include C(repository), C(repository_source_location), C(installed_date), C(published_date),
+        C(updated_date), C(package_management_provider), C(installed_location), C(tags), C(compatible_ps_editions),
+        C(license_uri), C(project_uri), C(icon_uri), C(release_notes), and C(dependencies).
+      - When C(include_properties) is specified without any repository properties, this is automatically set to C(true) for optimization.
+    type: bool
+    default: false
```
jborean93, line 50: Same here we need `version_added` for this new option

```suggestion
    default: false
    version_added: 3.3.0
```

jborean93 (changes requested):
Changes look good overall. I've fixed up the CI issues and triggered it to run again with your changes. Looks like there's a few sanity/linting issues around the changelog fragments. The fragments need to use `\n` and not `\r\n`, some longer lines need to be shorted and I typically do it by using `>-` in yaml like

```yaml
bugfixes:
  - >-
    component - some info
    that can span multiple lines -
    url
```

While not critical it would be great to see a test or two with the new options. I won't stop from merging this PR without those tests but keep in mind without them there's a risk that a future change breaks that logic and it goes unnoticed until after a release happens.

changelogs/fragments/win_psmodule_info_include_properties.yml
ShawnHardwick: Oof lol

ShawnHardwick:
All conversations addressed, tests added for new functionality. CICD pipeline is successful.

jborean93 (approved):
Looks great, thank you very much

